PR #6, opened 2018-05-18
Restructure plugin code, rename class and channel name, cleanup etc

MaikuB:
Note that the restructuring would mean that by importing one file would allow accessing all the models etc via one import. Did not do this in my own plugin as a I have platform specific data. See https://www.dartlang.org/tools/pub/package-layout#implementation-files for details


MaikuB:
Note that the restructuring would mean that by importing one file would allow accessing all the models etc via one import. Did not do this in my own plugin as a I have platform specific data. See https://www.dartlang.org/tools/pub/package-layout#implementation-files for details

device_calendar/lib/src/device_calendar.dart
```diff
@@ -1,30 +1,23 @@
...
-  static final CalendarsService _instance =
-      new CalendarsService._createInstance();
+class DeviceCalendarPlugin {
+  static const MethodChannel channel = const MethodChannel('plugins.builttoroam.com/device_calendar');
+
+  static final DeviceCalendarPlugin _instance =
+      new DeviceCalendarPlugin._createInstance();
 
-  factory CalendarsService() {
+  factory DeviceCalendarPlugin() {
```
mkieres: I don't get why you renamed this to be DeviceCalendar**Plugin**. I know how you like to follow the 'convention(s)', but none of the flutter plugins have the word plugin in it. I would remove it or literally name it _Calendar_

MaikuB: There is no convention and the naming wasn't picked based on convention. it's named that way because that's what the class represents since we're developing a <b>plugin</b>. You also said you were fine with the name. If anything, naming it calendar is worse cos we'll have a Calendar model class later

MaikuB: it's also not named DeviceCalendar as it'll read awkward when you ```new DeviceCalendar()```, which would translate to creating a new device calendar but you're creating an instance of the device calendar plugin

mkieres: I haven't heard of any plugin/nuget that is using a plugin word in their classes structures. More than anything, I'm surprised that you're not following these 'conventions' (i.e. official flutter plugins). Can stay as is.

mkieres: I get the point, although there's also a Connectivity plugin, where you create a new Connectivity(), but you're not connecting to anything. In that case I would lean more towards a name like CalendarService or DeviceCalendarService (or any other meaningful pre/postfix)

MaikuB: The other plugins aren't in the same scenario we are in where if we followed what they have, we'd end up with a DeviceCalendar object returning you a collection of Calendar objects. Helps avoid confusing and improves readability of code

PR #7, opened 2018-05-18
Return serialised calendar list from iOS to Dart and decode it

MaikuB:


device_calendar/lib/src/models/calendar.dart
```diff
@@ -1,12 +1,20 @@
...
   String id;
-
-  /// Calendar display name
   String name;
 
-  Calendar(this.id, {this.name});
+  Calendar({this.id, this.name});
+
+  Calendar.fromJson(Map<String, dynamic> json) {
+    id = json['id'];
```
mkieres: check for null - in the next PR

PR #9, opened 2018-05-21
Retrieving calendar events

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarService.kt
```diff
@@ -52,26 +61,104 @@ public class CalendarService : PluginRegistry.RequestPermissionsResultListener {
...
+
+        return null;
+    }
+
+
+    public fun retrieveEvents(calendarId: String) {
+        if (ensurePermissionsGranted()) {
+            val calendar = retrieveCelandar(calendarId);
+            if (calendar == null) {
+                _channelResult?.error("calendar_retrieval_failure", "Couldn't retrieve the Calendar with ID ${calendarId}", null);
```
MaikuB: make these private constants or whatever's closest to that in Kotlin?

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/common/Constants.kt
```diff
@@ -0,0 +1,38 @@
+package com.builttoroam.devicecalendar.common
+
+import android.provider.CalendarContract
+
+class Constants {
```
MaikuB: i think this should be public final class according to this article i saw before
https://blog.egorand.me/where-do-i-put-my-constants-in-kotlin/

mkieres: All classes of Kotlin are final by default

device_calendar/lib/src/models/event.dart
```diff
@@ -0,0 +1,27 @@
+part of device_calendar;
+
+class Event {
+  String title;
+
+  DateTime start;
+  DateTime end;
+
+  Event.fromJson(Map<String, dynamic> json) {
+    if (json == null) {
```
MaikuB: actually, i'm not sure this null check should be there as it is. if json was null, the error gets swallowed up and you end up with an event object (given this is a constructor) whose properties have default values. I'd suggest putting in an assert statement to check the json argument https://www.dartlang.org/guides/language/language-tour#assert

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/common/Constants.kt
MaikuB: oh actually i misread, it's the generated Java code. Seems the recommendation is to have a top level/global const but can look into that later

device_calendar/lib/src/models/event.dart
nickrandolph: assert won't work in production code though. Options are to return or to throw exception. You'll need to handle exceptions anyhow, so you might as well throw one

mkieres: "Assert statements have no effect in production code; they’re for development only" - I'm not truly sure what solution you're suggesting here.

Instead of returning the object with default values, maybe we should create a base object with 'isEmpty' flag and set it when json is null? I can't think of a better solution here.

MaikuB: @mkieres it's more to ensure that an error gets raised. logically shouldn't happen but if you were debugging and passed in a null then it breaks on assert statement.

@nickrandolph's point is valid though I do see a lot of assert statements being used in the official plugins. whether or not it's the best approach, i'm not sure...probably stick to what Nick is suggesting with letting it return an exception

nickrandolph: yeh the asserts are being used in a similar way to how we use Debug.WriteLine... they're for development only so they don't accidentally break stuff

PR #14, opened 2018-05-22
Adding create event functionality

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarService.kt
```diff
@@ -30,14 +31,18 @@ import com.builttoroam.devicecalendar.models.Calendar
...
 import com.google.gson.Gson
+import android.provider.CalendarContract.Events
+import android.content.ContentValues
+import java.util.*
 
 
 public class CalendarService : PluginRegistry.RequestPermissionsResultListener {
 
     private val REQUEST_CODE_RETRIEVE_CALENDARS = 0;
     private val REQUEST_CODE_RETRIEVE_EVENTS = 1;
```
nickrandolph: Can we make these dependent on each other eg
private val REQUEST_CODE_RETRIEVE_EVENTS = REQUEST_CODE_RETRIEVE_CALENDARS + 1;

mkieres: I mean, it doesn't matter what value it will have but sure

MaikuB: @nickrandolph shouldn't we prefer to make these a const or enum though? not sure what kind of behaviour your suggestion will have as well when it comes to Java code that gets generated. saw an article on how it translates the definition of constants https://blog.egorand.me/where-do-i-put-my-constants-in-kotlin/

nickrandolph: There's probably arguments for and against different structures and their relative perf (your link is very focused on how it translates to java rather than the intent of the language). See https://discuss.kotlinlang.org/t/constants-class/2589/2 and https://kotlinlang.org/docs/reference/object-declarations.html for more information on structuring const values
Also, I think making them derived values shouldn't preclude them from being constants. Definitely in C# you can do this and the compiler works out that they're still constants after compilation.

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
```diff
@@ -25,10 +26,14 @@ class DeviceCalendarPlugin() : MethodCallHandler {
     val RETRIEVE_CALENDARS_METHOD = "retrieveCalendars";
     val RETRIEVE_CALENDAR_EVENTS_METHOD = "retrieveEvents";
```
MaikuB: i think we should remove _CALENDAR_ from there since the Dart method is retrieveEvents not retrieveCalendarEvents

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
```diff
@@ -59,15 +64,28 @@ class DeviceCalendarPlugin() : MethodCallHandler {
 
         when (call.method) {
             RETRIEVE_CALENDARS_METHOD -> {
-                val calendars = _calendarService.retrieveCalendars();
+                _calendarService.retrieveCalendars();
             }
             RETRIEVE_CALENDAR_EVENTS_METHOD -> {
                 val calendarId = call.argument<String>(CALENDAR_ID_ARGUMENT);
-                if (calendarId?.isNullOrEmpty() ?: true) {
+                if (calendarId == null || calendarId.isEmpty()) {
```
MaikuB: you probably noted it already but will need to handle reading in the start and end date arguments with the changes i recently made

device_calendar/example/lib/main.dart
```diff
@@ -16,7 +16,7 @@ class _MyAppState extends State<MyApp> {
     return new MaterialApp(routes: {
       AppRoutes.calendars: (context) {
         return new CalendarsPage();
-      }
+      },
```
MaikuB: intentional trailing comma?

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
mkieres: Yep

device_calendar/example/lib/main.dart
mkieres: I had code there but then removed it. I will remove the comma

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
mkieres: Can do

PR #15, opened 2018-05-23
Updating query logic to filter through the calendar events with start and end date

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarService.kt
```diff
@@ -185,21 +192,29 @@ public class CalendarService : PluginRegistry.RequestPermissionsResultListener {
...
+        _calendarEventsStartDate = startDate;
+        _calendarEventsEndDate = endDate;
         if (ensurePermissionsGranted(REQUEST_CODE_RETRIEVE_EVENTS)) {
             val calendar = retrieveCalendar(calendarId, true);
             if (calendar == null) {
                 finishWithError(CALENDAR_RETRIEVAL_FAILURE, "Couldn't retrieve the Calendar with ID ${calendarId}");
                 return;
             }
+            if (startDate > endDate) {
+                finishWithError(EVENTS_RETRIEVAL_FAILURE, EVENTS_START_DATE_LARGER_THAN_END_DATE_MESSAGE);
```
MaikuB: hmm good point, though i think maybe we should catch this at the Dart level and throw an exception there as it'll likely be more performant that way. can be done later though, just a thought

PR #16, opened 2018-05-23
Adding ability to update an event. Fixing some of the date & time logic in the example

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarService.kt
```diff
@@ -270,12 +270,17 @@ public class CalendarService : PluginRegistry.RequestPermissionsResultListener {
...
+                    eventId = java.lang.Long.parseLong(uri?.getLastPathSegment());
+                } else {
+                    contentResolver?.update(ContentUris.withAppendedId(CalendarContract.Events.CONTENT_URI, eventId), values, null, null);
+                }
 
-                // get the event ID that is the last element in the Uri
-                val eventID = java.lang.Long.parseLong(uri?.getLastPathSegment());
 
-                finishWithSuccess(eventID.toString());
+                finishWithSuccess(eventId?.toString());
```
MaikuB: should we be checking that eventId isn't null at this level instead of swallowing it or do you check at the Flutter/Dart layer?

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
```diff
@@ -27,7 +27,7 @@ class DeviceCalendarPlugin() : MethodCallHandler {
     val RETRIEVE_CALENDARS_METHOD = "retrieveCalendars";
     val RETRIEVE_EVENTS_METHOD = "retrieveEvents";
     val DELETE_EVENT_METHOD = "deleteEvent";
-    val CREATE_EVENT_METHOD = "createEvent";
+    val CREATE_OR_UPDATE_EVENT_METHOD = "createOrUpdateEvent";
```
MaikuB: need to update the request code in the calendar service class as well

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarService.kt
mkieres: I thought I am handling this on the dart level. Turns out, I'm not. Will make sure that we are handling this scenario. Good catch

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
mkieres: Yep, forgot about it

PR #18, opened 2018-05-24
Add ability to add or update an event in iOS

MaikuB:
Also fixes example code to cater for scenarios where events might have missing info like null description etc

### DIFF
--- a/device_calendar/example/lib/presentation/event_item.dart
+++ b/device_calendar/example/lib/presentation/event_item.dart
@@ -29,8 +29,8 @@ class EventItem extends StatelessWidget {
               child: new FlutterLogo(),
             ),
             new ListTile(
-                title: new Text(_calendarEvent.title),
-                subtitle: new Text(_calendarEvent.description)),
+                title: new Text(_calendarEvent.title ?? ''),
+                subtitle: new Text(_calendarEvent.description ?? '')),
             new Container(
               padding: new EdgeInsets.symmetric(horizontal: 16.0),
               child: new Column(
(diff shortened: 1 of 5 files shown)


device_calendar/example/lib/presentation/event_item.dart
```diff
@@ -29,8 +29,8 @@ class EventItem extends StatelessWidget {
               child: new FlutterLogo(),
             ),
             new ListTile(
-                title: new Text(_calendarEvent.title),
-                subtitle: new Text(_calendarEvent.description)),
+                title: new Text(_calendarEvent.title ?? ''),
```
mkieres: Wouldn't this just work? As in, when null is provided then the text is basically empty?

device_calendar/lib/src/common/constants.dart
```diff
@@ -2,5 +2,6 @@ class Constants {
   static const String fromJsonMapIsNull = 'The json object is null';
 
   static const String invalidArgument = 'Provided arguments are invalid';
-  static const String createOrUpdateEventArgumentRequirements = 'To create or update an event you must provide calendar id, event with title, start date and end date (where start date must be before end date)';
+  static const String createOrUpdateEventArgumentRequirements =
```
mkieres: Is this a correct formatting?

MaikuB: yes via flutter format

mkieres: Sorry, I thought there's an extra line space between the variable name and its value.

You should be careful with flutter format though, as sometimes it breaks the logic of the app - situation that I had with chaining extension methods

PR #24, opened 2018-05-30
Creating request permission and has permissions methods

mkieres:


device_calendar.code-workspace
```diff
@@ -1,11 +1,11 @@
 {
 	"folders": [
-		{
-			"path": "device_calendar"
-		},
 		{
 			"path": "device_calendar/example",
```
MaikuB: oops i'll need to delete this/add to .gitignore

mkieres: Yeah, I was thinking that as well

mkieres: Dunno how to do it...

PR #28, opened 2018-05-30
Returning event attendees

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/models/Attendee.kt
```diff
@@ -0,0 +1,8 @@
+package com.builttoroam.devicecalendar.models
+
+class Attendee(val name: String) {
+    var id: Long = -1
```
MaikuB: FYI, we only need the name at this stage as other properties like email can't be retrieved from iOS or that there are enums like attendance status that don't quite line up on both platforms for us to provide an abstraction

mkieres: Yep, I think I will leave it there anyhow. 

Would this help (https://stackoverflow.com/a/17222036/510627) to retrieve the email address for the iOS implementation? 

I'm not entirely sure what are you referring to as status (as the names are different across platforms) and I haven't looked much into iOS EKParticipant model but at least for what I implemented on Droid (if user is required to be present at the event) there's a a way to retrieve such info with https://developer.apple.com/documentation/eventkit/ekparticipantrole

PR #29, opened 2018-05-31
Address codefactor issues

MaikuB:


device_calendar/pubspec.yaml
```diff
@@ -1,53 +1,19 @@
 name: device_calendar
-description: A new flutter plugin project.
+description: A cross platform plugin for displaying local notifications.
```
mkieres: I like how you copy pasta without checking ;]

PR #30, opened 2018-05-31
Handling platform errors. Unifying droid and iOS error codes

mkieres:
You might want to add NOT_ALLOWED error code to iOS and handle the isReadOnly flag for the calendar, when user tries to do a CRUD operation on the calendar

### DIFF
--- a/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
+++ b/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
@@ -357,7 +357,7 @@ public class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener 
 
             val contentResolver: ContentResolver? = _context?.getContentResolver()
             if (isRecurringEvent(eventIdNumber, contentResolver)) {
-                finishWithError(NOT_FOUND, DELETING_RECURRING_EVENT_NOT_SUPPORTED_MESSAGE, pendingChannelResult)
+                finishWithError(NOT_ALLOWED, DELETING_RECURRING_EVENT_NOT_SUPPORTED_MESSAGE, pendingChannelResult)
                 return
             }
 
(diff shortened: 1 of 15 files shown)


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -348,12 +341,12 @@ public class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener
...
             var existingCal = retrieveCalendar(calendarId, pendingChannelResult, true)
             if (existingCal == null) {
-                finishWithError(CALENDAR_RETRIEVAL_FAILURE, "Couldn't retrieve the Calendar with ID ${calendarId}", pendingChannelResult)
+                finishWithError(NOT_FOUND, "Couldn't retrieve the Calendar with ID ${calendarId}", pendingChannelResult)
                 return
             }
 
             if (existingCal.isReadyOnly) {
-                finishWithError(CALENDAR_IS_READ_ONLY, "Calendar with ID ${calendarId} is read only", pendingChannelResult)
+                finishWithError(NOT_ALLOWED, "Calendar with ID ${calendarId} is read only", pendingChannelResult)
```
MaikuB: 405 represents a method that doesn't exist. I reckon it should be 403. i'll take a look at the iOS code and see will add that in

mkieres: Where do you get the info about the statuses? Wikipedia says "405 Method Not Allowed
A request method is not supported for the requested resource". 

403 is more related to missing permissions - I would say

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -365,7 +358,7 @@ public class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener
 
             val contentResolver: ContentResolver? = _context?.getContentResolver()
             if (isRecurringEvent(eventIdNumber, contentResolver)) {
-                finishWithError(DELETING_RECURRING_EVENT_NOT_SUPPORTED, DELETING_RECURRING_EVENT_NOT_SUPPORTED_MESSAGE, pendingChannelResult)
+                finishWithError(NOT_FOUND, DELETING_RECURRING_EVENT_NOT_SUPPORTED_MESSAGE, pendingChannelResult)
```
MaikuB: i'd say the 405 is appropriate here

mkieres: I can agree to that

device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
MaikuB: hmm i guess could argue both ways as read-only means it could just be only by the OS itself or maybe it's another person's calendar (assuming it's possible to get other users calendars, i don't know if that can be done). guess leave it as it is then

PR #39, opened 2018-06-05
Comments for review and discussion

nickrandolph:


device_calendar/lib/src/device_calendar.dart
```diff
@@ -74,6 +81,8 @@ class DeviceCalendarPlugin {
       res.errorMessages.add(
           "[${ErrorCodes.invalidArguments}] ${ErrorMessages.invalidMissingCalendarId}");
     }
+
+    // TODO: Extend capability to handle null start or null end (eg all events after a certain date (null end date) or all events prior to a certain date (null start date))
```
MaikuB: iOS requires both start and end to be specified if dates are used and this is the only predicate that they allow using
https://developer.apple.com/documentation/eventkit/ekeventstore/1507479-predicateforeventswithstartdate?language=objc

retrieve by id is done using this https://developer.apple.com/documentation/eventkit/ekeventstore/1507490-eventwithidentifier?language=objc

device_calendar/lib/src/common/error_codes.dart
```diff
@@ -1,3 +1,4 @@
 class ErrorCodes {
   static const int invalidArguments = 400;
+  // TODO: Add error code 500 for when we get an error back from the device level
```
mkieres: Let's add platform exception (500)

device_calendar/lib/src/device_calendar.dart
```diff
@@ -16,21 +16,26 @@ class DeviceCalendarPlugin {
   DeviceCalendarPlugin._createInstance();
 
   /// Requests permissions to modify the calendars on the device
+  /// TODO: add comment about return value
```
mkieres: Investigate on how to create comments on the code - input parameters and return values

device_calendar/lib/src/models/result.dart
```diff
@@ -5,5 +5,11 @@ class Result<T> {
   T data;
   List<String> errorMessages = new List<String>();
 
+    // TODO: Change to allow constructor parameter
   Result([this.data]);
+
```
mkieres: 1. Make IsSuccess flag dependant on Data not being null and error messages being empty
2. Change constructor to be parameter free

device_calendar/lib/src/device_calendar.dart
```diff
@@ -16,21 +16,26 @@ class DeviceCalendarPlugin {
...
 
     try {
       res.data = await channel.invokeMethod('requestPermissions');
+      // TODO: Remove calls to set isSuccess to true
       res.isSuccess = true;
-    } on PlatformException catch (e) {
+    } on PlatformException catch (e) { 
+      // TODO: Change parse method to return a new Result that can be returned immediately     
       _parsePlatformExceptionAndUpdateResult<bool>(e, res);
+      // TODO: Move print into the parse method
```
mkieres: Move this to the parsing method

device_calendar/lib/src/device_calendar.dart
mkieres: Pushing this to the next iteration

device_calendar/lib/src/device_calendar.dart
```diff
@@ -85,6 +94,7 @@ class DeviceCalendarPlugin {
           "[${ErrorCodes.invalidArguments}] ${ErrorMessages.invalidRetrieveEventsParams}");
     }
 
+    // TODO: Change this to check Success
```
mkieres: Doing it!

device_calendar/lib/src/device_calendar.dart
```diff
@@ -168,9 +180,11 @@ class DeviceCalendarPlugin {
     return res;
   }
 
+  // TODO: Change this to return a new Result based on the parsed exception
```
mkieres: Handle scenario when the exception is null - create a generic message about 'unknown' platform error

PR #50, opened 2018-06-08
Creating a brand new example to appear in the pub dart Example tab. Moving existing sample to example_app folder. Updating changelog

mkieres:


device_calendar/example/lib/main.dart
```diff
@@ -13,10 +10,16 @@ class MyApp extends StatefulWidget {
...
   @override
   Widget build(BuildContext context) {
-    return new MaterialApp(routes: {
-      AppRoutes.calendars: (context) {
-        return new CalendarsPage();
-      }
-    });
+    return new MaterialApp(
+      title: 'Device Calendar Example',
+      home: new Scaffold(
```
nickrandolph: We still need to add some actual examples showing using the calendar in here, right?

mkieres: Yep, I just wanted to check this in, so that we can all agree that this is how the project/example structure should look like.

PR #55, opened 2018-06-14
Fixing an issue with retrieving events by id

mkieres:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -246,24 +247,29 @@ public class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener
     }
 
     @SuppressLint("MissingPermission")
-    public fun retrieveEvents(calendarId: String, startDate: Long, endDate: Long, eventIds: List<String>, pendingChannelResult: MethodChannel.Result) {
+    public fun retrieveEvents(calendarId: String, startDate: Long?, endDate: Long?, eventIds: List<String>, pendingChannelResult: MethodChannel.Result) {
+        if (startDate == null && endDate == null && eventIds.isEmpty()) {
```
nickrandolph: I'm not familiar with KT but do we need to null check eventIds before calling isEmpty?

mkieres: No, that's the thing about KT, that when you don't say it can be null (e.g. Long?) then you basically can't. You need to pass something that's not null

nickrandolph: ok cool, just checking that applied to more complex objects like a List.

PR #81, opened 2019-02-26
Correctly increment minor version instead of patch version

MaikuB:


MaikuB:
true, will correct wording

PR #85, opened 2019-03-01
merge latest releases to master

MaikuB:


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -90,7 +90,7 @@ public class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener
             // No exceptions, this should not happen, ever
             //
             // There's nothing that can be done at this stage, besides finishing gracefully
-            return true
+            return false
```
nickrandolph: The comment implies return value should be true? Can we provide a better explanation

MaikuB: noted. will update comments in next release/PR

PR #87, opened 2019-05-09
WIP Updated attendee model

alexandrucozmanca:


MaikuB:
we don't currently allow for specifying this when creating events, which is why they can't be specified via the constructor at the moment. allowing them to be specified in the constructor could lead other devs to think it can be passed through to the native platform's calendar APIs

MaikuB:
There are some significant differences between the platforms on what information is available for attendees. As such, will be looking to close this PR as it's more Android-specific

PR #98, opened 2019-06-06
Compatibility fixes

duzenko:
I suspected `calendar(withIdentifier:` to be the culprit but it turns out that the cast fix resolves the original problem
https://github.com/builttoroam/flutter_plugins/issues/94


duzenko:
The only change in the example app was build for arm-32
I can do that on the other example as well if that's what you mean

MaikuB:
Yep that's what I meant

PR #103, opened 2019-07-20
iOS location and location coordinates support for create / update event

oli06:
Hi,
I added iOS location support for creation of events (Showing the Location String inside the event)
Coordinates (String, lat and long separeted by ',') are also provided to show a map inside the iOS Calendar Event View.


MaikuB:
are you able to add changes to support this on Android?

oli06:
Hi @MaikuB 
as far as I know, there is no way provided to store latitude and longitude inside an event object on Android. I'm not very familiar with Kotlin / Android Development but as this page shows https://developer.android.com/reference/android/provider/CalendarContract.EventsColumns.html#EVENT_LOCATION there is just a normal String location. 

Besides that, the iOS Calendar provides functionality to show a map inside an event, if coordinates are provided (example: https://i.stack.imgur.com/6Y620.png). You can click on the map to open the route in for example Apple Maps.

MaikuB:
think the options here then are to either start having platform-specific event details in a separate class or have a way of querying what details can be entered. what's the reasoning behind making longitude and latitude combined as a string?

also should point there are conflicts in the code at the moment

MaikuB:
Took another look at the coordinates that got added. It's not actually so that iOS can show a map even though that is the effect but if you look at the documentation provided by Apple, `structuredLocation` is done to setup a geofence to trigger a reminder based on the user's location. As such, I think this requires thought on how to get this supported so I've removed the code around this for now

PR #127, opened 2019-11-07
Returned an isDefault property for calendar

jbuda:


nickrandolph:
@jbuda do you know whether there's a similar property on Android? We'd like to maintain consistency across the supported platforms (atm iOS and Android).

jbuda:
@nickrandolph Off the top of my head i can't remember. However, I can take a look when i get round to implementing the Calendar into the Android version.

nickrandolph:
@Brett09 I think we can use this field on Android: https://developer.android.com/reference/android/provider/CalendarContract.CalendarColumns.html#IS_PRIMARY

Brett09:
@jbuda thanks for providing iOS support. I'll be merging this with Android support into my branch then create a new PR, hence I'm going to close this.

PR #136, opened 2020-01-08
Brettl/add url

Brett09:
- Custom URL added for calendar events.
- Updated iOS cocoapods to latest (1.8.4)


device_calendar/ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -415,6 +419,10 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin {
             ekEvent!.endDate = endDate
             ekEvent!.calendar = ekCalendar!
             ekEvent!.location = location
+
+            let iosUrl = URL(string: url ?? "")
```
nickrandolph: if url is null/empty string, we shouldn't be creating the URL object or setting the url property

device_calendar/lib/src/models/event.dart
```diff
@@ -31,6 +31,9 @@ class Event {
   /// The location of this event
   String location;
 
+  /// An URL for this event
+  String url;
```
nickrandolph: Update this to use the Uri class so that we can validate that the url entered conforms to a url (this is already done by iOS as it requires a Url but Android exposes this as a string)

PR #144, opened 2020-01-20
Attendee ui

Brett09:
**Issue:** #141

**Changes**
- Reads if an attendee is an organiser
- Attendee UI update
- Gradle/Swift update

**Possible Breaking Changes**
1. In `attendee.dart` file, `isRequired` boolean property has been replaced to `AttendeeRole` enum
2. New argument added for the `AttendeeRole` property

### DIFF
--- a/device_calendar/example/lib/presentation/pages/event_attendees.dart
+++ b/device_calendar/example/lib/presentation/pages/event_attendees.dart
@@ -81,7 +81,7 @@ class _EventAttendeesPageState extends State<EventAttendeesPage> {
                       name: _nameController.text,
                       emailAddress: _emailAddressController.text,
                       role: _role,
-                      isOrganiser: false,
+                      //isOrganiser: false,
                     );
 
                   _emailAddressController.clear();
--- a/device_calendar/example/lib/presentation/pages/event_attendees.dart
+++ b/device_calendar/example/lib/presentation/pages/event_attendees.dart
@@ -2,11 +2,12 @@ import 'package:flutter/material.dart';
 import 'package:device_calendar/device_calendar.dart';
 
 class EventAttendeesPage extends StatefulWidget {
-  const EventAttendeesPage({Key key}) : super(key: key);
+  final Attendee attendee;
+  const EventAttendeesPage({Key key, this.attendee}) : super(key: key);
 
   @override
   _EventAttendeesPageState createState() =>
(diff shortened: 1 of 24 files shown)


device_calendar/example/lib/presentation/pages/event_attendees.dart
```diff
@@ -2,25 +2,33 @@ import 'package:flutter/material.dart';
 import 'package:device_calendar/device_calendar.dart';
 
 class EventAttendeesPage extends StatefulWidget {
```
nickrandolph: Change name of page to EventAttendeePage <-- remove the S on Attendee

device_calendar/lib/src/common/calendar_enums.dart
```diff
@@ -134,4 +140,12 @@ extension IntExtensions on int {
   DayOfWeek get getDayOfWeekEnumValue => _getDayOfWeekEnumValue(this);
   MonthOfYear get getMonthOfYearEnumValue => _getMonthOfYearEnumValue(this);
   WeekNumber get getWeekNumberEnumValue => _getWeekNumberEnumValue(this);
+}
+
+extension ObjectExtensions on Object {
```
nickrandolph: Can we make this explicit to the type of enum we want to convert to string (eg https://stackoverflow.com/questions/38908285/add-methods-or-values-to-enum-in-dart)

device_calendar/lib/src/models/platform_specifics/ios/attendee_details.dart
```diff
@@ -16,17 +17,15 @@ class IosAttendeeDetails {
...
     if (json['attendanceStatus'] != null && json['attendanceStatus'] is int) {
       _attendanceStatus = IosAttendanceStatus.values[json['attendanceStatus']];
     }
   }
 
   Map<String, dynamic> toJson() {
-    return <String, dynamic>{
-      'role': role?.index,
-    };
+    return <String, dynamic>{ 'role': role.index };
```
nickrandolph: Is role guarranteed to not be null? Might be safe to keep the role?.index just to be safe.

Brett09: It is guaranteed not to be null (always default to 'None' role), but I can add the null check in case.

PR #148, opened 2020-01-22
Day of week group

Brett09:
**Issue:** #147 

**Changes**
* New `DayOfWeekGroup` enum consisting Weekday, Weekend and Alldays
* Using `getDays` extension to get a list of `DayOfWeek` enums explicitly
* Added to the example app
* Updated example readme and changelog


device_calendar/example/lib/presentation/pages/calendar_event.dart
```diff
@@ -688,6 +711,45 @@ class _CalendarEventPageState extends State<CalendarEventPage> {
     }
   }
 
+  void _updateDaysOfWeek(DayOfWeekGroup selectedGroup, bool isSelected) {
+    var days = selectedGroup.getDays;
+
+    if (isSelected) {
+      _daysOfWeek.addAll(days.where((a) => _daysOfWeek.every((b) => a != b)));
```
nickrandolph: what happens if _daysOfWeek already has values that are not in days - do we need to remove them?

Brett09: Nothing, that `allAll` condition will only add whatever it's not in `_daysOfWeek` list.
i.e. If `_daysOfWeek` already have Monday, Tuesday and Wednesday, choosing `DayOfWeekGroup.Weekday` will only add Thursday and Friday.

Brett09:
I'm rebasing with `develop` to resolve the conflicts now, will be committed soon.

device_calendar/example/lib/presentation/pages/calendar_event.dart
nickrandolph: but what happens if _daysOfWeek has sat or sun... shouldn't they be removed if Weekday is specified?

Brett09: ok how about adding `None` and `Custom` as you've mentioned before and have `DayOfTheWeekGroup` as radio buttons in the example app?

This way if `_daysOfWeek` has Monday to Saturday, `Custom` will be chosen, choosing `Weekday` will just remove Saturday.

Brett09: @nickrandolph I've changed it to be a radio button with `None` and `Custom` added to the enum

PR #157, opened 2020-01-28
fixed nullable allDay error

FedericoMatera:


nickrandolph:
@FedericoMatera I've changed the target branch to be master. Would you be able to do a rebase so that we're only seeing your change in the commit history. This way we can hotfix your bug fix into production without pulling the latest changes from develop too.

PR #160, opened 2020-01-30
Readonly_noadd

Brett09:
Quick UI fix to hide an add event button for read-only calendars


device_calendar/example/lib/presentation/pages/calendar_events.dart
```diff
@@ -73,8 +79,11 @@ class _CalendarEventsPageState extends State<CalendarEventsPage> {
           }
         },
         child: Icon(Icons.add),
-      ),
-    );
+      );
+    }
+    else {
+      return Container();
```
nickrandolph: Do we need to return a container here, or can we just return null?

Brett09: @nickrandolph updated

PR #176, opened 2020-03-05
Build Pipelines

Brett09:
Issue added for this PR: #177


azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+        - task: CmdLine@2
+          displayName: 'Publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --force'
+            
+  # ----------- CD Production -----------
+  - stage: ReleaseProd
+    dependsOn: Build
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/master'))
```
mkieres: Should we be releasing from `release` branch instead?

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+  # See code in the Flutter Build task https://github.com/aloisdeniel/vsts-flutter-tasks/blob/master/tasks/build/index.ts
+  flutterExecPath: $(FlutterToolPath)/flutter.bat
+
+trigger:
+  batch: 'true'
+  branches:
+    include:
+      - develop
+      - release
+      - master
```
mkieres: Do we want to trigger this CI / CD pipeline from `master` branch?

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+
+        - task: FlutterBuild@0
+          displayName: 'Flutter build - iOS'
+          inputs:
+            target: 'ios'
+            projectDirectory: 'device_calendar/example'
+            iosCodesign: false
+            
+  # -----------CD Pre-release -----------
+  - stage: ReleaseBeta
```
nickrandolph: Rename to "DevelopmentRelease"

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+        - task: FlutterInstall@0
+          displayName: 'Flutter install'
+          inputs:
+            channel: 'stable'
+            version: 'latest'
+
+        - task: FlutterBuild@0
+          displayName: 'Flutter build - Android'
+          inputs:
+            target: 'apk'
```
mkieres: I wonder if it would be better if we try building aab (`appbundle`) instead. Don't think it matters from the build verifications perspective but I guess aab is the new "way to go" so maybe we should try building it?

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+            target: 'ios'
+            projectDirectory: 'device_calendar/example'
+            iosCodesign: false
+            
+  # -----------CD Pre-release -----------
+  - stage: ReleaseBeta
+    dependsOn: Build
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/release'))
+    jobs:
+     - job: Release
```
nickrandolph: Rename to "ReleaseAsDevelopment

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+            script: '$(flutterExecPath) pub publish --dry-run'
+
+        - task: CmdLine@2
+          displayName: 'Publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --force'
+            
+  # ----------- CD Production -----------
+  - stage: ReleaseProd
```
nickrandolph: Rename to just "Release"

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+          displayName: 'Flutter install'
+          inputs:
+            channel: 'stable'
+            version: 'latest'
+
+        - task: CmdLine@2
+          displayName: 'Dry run publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --dry-run'
```
mkieres: I might be missing something but what's the reasoning for doing the `dry-run` before actual publish? Wouldn't the publish itself be sufficient? If it fails it won't release anything anyways, right?

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+
+        - task: CmdLine@2
+          displayName: 'Publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --force'
+
+        - task: PowerShell@2
```
nickrandolph: We should do this before doing the publish.
We should also query github to see if the tag exists already - if it does, the build should fail.

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+            targetType: 'inline'
+            script: |
+              [string] $version = select-string -path "$(Build.SourcesDirectory)/device_calendar/pubspec.yaml" -pattern "(?<=version: ).*" | %{ $_.Matches[0].Value }
+              
+              Write-Host "##vso[task.setvariable variable=currentVersion]$version"
+
+        - task: GitHubRelease@1
+          displayName: 'Tag a release to GitHub'
+          inputs:
+            gitHubConnection: 'github.com_Brett09'
```
mkieres: Shouldn't we have like a service account for this task, instead of using a personal one?

nickrandolph: We need to move this github connection to be one using the built to roam org or devops account - can you walk me through creating this and then we can update this.

azure-pipelines.yml
```diff
@@ -0,0 +1,164 @@
...
+            repositoryName: 'builttoroam/flutter_plugins'
+            action: 'create'
+            target: 'master'
+            tagSource: 'userSpecifiedTag'
+            tag: v$(currentVersion)
+            title: v$(currentVersion)
```
nickrandolph: We should pull out the text from version line in the change log. In the past this has just been the date of the release but I think going forward this should be a single line summary of the change eg Added Timezone Support

Brett09: What if there are multiple items (eg. multiple new feature added)? As an example you can see the v3.1.0 changelog in `develop` branch, there are a lot of changes.

Maybe from now on we should have a habbit of having a summary in the first line of each version in the changelog?

azure-pipelines.yml
nickrandolph: All three - the first step is a CI so should run on all

azure-pipelines.yml
nickrandolph: correct that's what I was thinking - just have it on same line as the version number eg a title

azure-pipelines.yml
nickrandolph: long conversation with brett on this one and for the moment this is what we agreed - would be good to get your input on this though as we can evolve this if we're not comfortable with this approach

azure-pipelines.yml
Brett09: Nick has mentioned it already and will go through that with me

azure-pipelines.yml
Brett09: I'm using `pub publish --force` for the actual publish which bypasses all warnings (and skip the input for confirming the publish), but `--dry-run` returns an error if we get a warning. This enables to fix all warnings before doing an actual publish.

azure-pipelines.yml
Brett09: We usuall have a release date on the same line of the version (i.e. for version 3.0.0 we have: `3.0.0 21 Janurary 2020`)

Do you still want to put the date and put a summary after? or forget about the date?

azure-pipelines.yml
Brett09: Will give it a go!

azure-pipelines.yml
Brett09: so our decisions were:

- merge to `develop` - trigger CI
- merge to `release` - trigger CI and CD pre-release
- merge to `master` - trigger CI and CD production

azure-pipelines.yml
Brett09: Updated to the PAT now

azure-pipelines.yml
Brett09: @nickrandolph I've moved to run the github release before publish, but need to find out how i can fail the build if a tag exists already (current task doesn't support this, it just creates for the existing tag)

Brett09: @nickrandolph, used the new task and tested an actual release with an existing tag and it returned an error

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+            script: |
+              [string] $summary = select-string -path "$(Build.SourcesDirectory)/device_calendar/CHANGELOG.md" -pattern "($([regex]::escape($env:currentVersion))).*" | %{ $_.Matches[0].Value }
+              
+              Write-Host "##vso[task.setvariable variable=versionSummary]$summary"
+
+        - task: GitHubReleasePublish@1
+          displayName: 'Tag a release to GitHub'
+          inputs:
+            githubEndpoint: 'GitHub Tagging'
+            manuallySetRepository: false
```
mkieres: Is this condition set properly? I mean it would indicate that you don't want to set repo manually and then in the line below you're setting it. I don't know enough about this task so might be comletely off

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+
+        - task: CmdLine@2
+          displayName: 'Publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --force'
+
+        - task: GitHubRelease@1
+          displayName: 'Remove GitHub tag when publish fails'
+          condition: failed()
```
mkieres: I don't think it's a big deal but the [`failed()` expression](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#failed) will evaluate to `True` whenever any previous task will fail, meaning that even when any PowerShell script fails this will run (at least that's my reading of it). It's not ideal and maybe we could find a way to limit this run to only run when the `Publish` or `Dry-Run` task fails? Basically anything after setting up the tag taks

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+            target: 'ios'
+            projectDirectory: 'device_calendar/example'
+            iosCodesign: false
+            
+  # -----------CD Pre-release -----------
+  - stage: DevelopmentRelease
+    dependsOn: Build
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/release'), eq(variables['Build.SourceBranch'], 'refs/hotfix'))
+    jobs:
+     - job: ReleaseAsDevelopmentVersion
```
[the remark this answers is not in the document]
Brett09: That's a recommendation from @nickrandolph

azure-pipelines.yml
Brett09: Yup tested (as creating drafts) and it works fine.

azure-pipelines.yml
Brett09: You're right and I'm aware of this and couldn't find a way to do that on the task level.

Brett09: I'll have more look but cannot guarantee

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+
+              $newChangelogContent | Set-Content -Path "$(Build.SourcesDirectory)/device_calendar/CHANGELOG.md"
+
+        - task: CmdLine@2
+          displayName: 'Dry run publish'
+          inputs:
+            workingDirectory: '$(Build.SourcesDirectory)/device_calendar'
+            script: '$(flutterExecPath) pub publish --dry-run'
+
+        - task: CmdLine@2
```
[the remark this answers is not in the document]
Brett09: This is how it should work:
1. Merging a local branch to `hotfix`
    - Only run `--dry-run` and will not get a warning regardless of pubspec
2. Merging `hotfix` into `master`
    - This is just merging from a branch to another, so I reckon the developer & reviewer should make sure the version is a hotfix version?

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+        vmImage: 'windows-latest'
+       steps:
+        - task: FlutterInstall@0
+          displayName: 'Flutter install'
+          inputs:
+            channel: 'stable'
+            version: 'latest'
+
+        - task: PowerShell@2
+          condition: eq(variables['Build.SourceBranch'], 'refs/release')
```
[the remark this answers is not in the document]
Brett09: Added 'refs/release' to a global variable as it's repeated several times, but do you recommend to use all strings to be a global variable?

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+          displayName: 'Pre-release versioning - pubspec.yaml'
+          inputs:
+            targetType: 'inline'
+            script: |
+              [string] $version = select-string -path "$(Build.SourcesDirectory)/device_calendar/pubspec.yaml" -pattern "(?<=version: ).*" | %{ $_.Matches[0].Value }
+              Write-Host "##vso[task.setvariable variable=currentVersion]$version"
+
+              [string] $versionRegex = '(?<=version: ).*'
```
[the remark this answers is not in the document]
Brett09: Redundant. removed

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+              [string] $pubspecContent = Get-Content -Path "$(Build.SourcesDirectory)/device_calendar/pubspec.yaml" -Raw
```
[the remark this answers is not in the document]
Brett09: Redundant. removed

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+
+              Write-Host " (i) Current version: $version"
+              Write-Host " (i) Original pubspec.yaml content: $pubspecContent"
+
+              [bool] $hasMatches = $pubspecContent -match $versionRegex
```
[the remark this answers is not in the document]
Brett09: Thanks for that. Removed the `-match-` script

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+          inputs:
+            channel: 'stable'
+            version: 'latest'
+
+        - task: PowerShell@2
+          displayName: 'Get version from pubspec.yaml'
+          inputs:
+            targetType: 'inline'
+            script: |
+              [string] $version = select-string -path "$(Build.SourcesDirectory)/device_calendar/pubspec.yaml" -pattern "(?<=version: ).*" | %{ $_.Matches[0].Value }
```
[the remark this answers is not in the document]
Brett09: Moved to a global variable

azure-pipelines.yml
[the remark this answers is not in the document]
Brett09: Wait, the build fails if I use a global variable inside a condition... Maybe I should just use a string for those branch names in the condition?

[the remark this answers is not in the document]
mkieres: I'm not really sure how to deal with variables in expressions (maybe not even possible @nickrandolph ?). Anyhow, you could potentially do it slightly differently and evaluate the entire expression and assign it to a variable, see top of the docs here https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops

azure-pipelines.yml
mkieres: You might be right that there's no way of doing it (@nickrandolph ?), maybe we should then split this into multiple jobs? That way you could use `dependsOn` with the `condition: failed()`?

azure-pipelines.yml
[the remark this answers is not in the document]
mkieres: I would still drop it :P - not too fussed about it though. Leave as is

azure-pipelines.yml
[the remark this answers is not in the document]
mkieres: Make sense. Thanks for explanation. I forgot that the publish for `hotfix` only happens from the `master` and yes, you're right we will need to be careful when merging into master with versioning. 

Potentially for merges from `hotfix` to `master` we could have some sort of a check in terms of versioning in the `pubspec`, to ensure we're not going to bump the version incorrectly. Probalby something to think about for later

azure-pipelines.yml
```diff
@@ -7,6 +7,8 @@ variables:
   # The $(FlutterToolPath) variable is generated by the FlutterInstall task 
   # See code in the Flutter Build task https://github.com/aloisdeniel/vsts-flutter-tasks/blob/master/tasks/build/index.ts
   flutterExecPath: $(FlutterToolPath)/flutter.bat
+  versionGetRegex: '(?<=version: ).*'
```
mkieres: I would probably drop `Get` from the name and potentially rename it to `versionNumberRegex` - feel free to adjust accordingly (i.e. if it's not actually a number)

azure-pipelines.yml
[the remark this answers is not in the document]
mkieres: I wouldn't make all of the strings global but the ones that are used across different stages / jobs. As a general rule of thumb you should try to move string literalls and "random" numbers in code to a const / variable to allow for a re-use but even more importantly to have them named, which will make them easier to read and understand the purpose.

azure-pipelines.yml
```diff
@@ -63,32 +65,27 @@ stages:
...
 
               Write-Host " (i) Current version: $version"
               Write-Host " (i) Original pubspec.yaml content: $pubspecContent"
-
-              [bool] $hasMatches = $pubspecContent -match $versionRegex
-
-              Write-Host " (i) Found matches to replace version? $hasMatches"
               
-              [string] $newPubspecContent = $pubspecContent -replace $versionRegex, "$version-dev.$(Build.BuildID)"
+              [string] $newPubspecContent = $pubspecContent -replace $version, "$version-dev.$(Build.BuildID)"
```
mkieres: I think this shouldn't be changed?

azure-pipelines.yml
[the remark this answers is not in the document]
mkieres: I would probably leave that code in (move it around though), as it might be helpful when reading logs to see if there were any matches or not. If you remove it then in case there are no matches you will get an exception message ( this will throw an exception `$_.Matches[0].Value`) which might not necessarily be very descriptive and harder to debug.

azure-pipelines.yml
Brett09: I tried `dependsOn` but cannot use it for tasks level (I guess `dependsOn` is available for stage level only)

azure-pipelines.yml
Brett09: Either way works as both variable returns the same value, are you happy if I leave it as it is?

azure-pipelines.yml
Brett09: Updated to your preference ;)

azure-pipelines.yml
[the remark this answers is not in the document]
Brett09: Why not, updated ;)

azure-pipelines.yml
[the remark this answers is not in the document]
Brett09: I've reverted for now

azure-pipelines.yml
mkieres: Hmm, are we using regex in other places with similar code? If so I would stick to regex for consistency.

azure-pipelines.yml
mkieres: You could use `dependsOn` at a job level as well. That was my suggestion to potentially split up these task into different jobs and make them depend on each other. Have a think about this - this is definitely not a blocker for this PR and can be done at later stage.

azure-pipelines.yml
mkieres: Hehehe, thanks. I know I can be very peculiar about naming of variables, classes etc., which may seem very picky - I apologise. The real reason behind my suggestions for changes in such cases is the fact that naming things in code is very important. Even though you think it might be irrelevant, the readability of code is a key factor to a good and clean code. Also it's really hard, to name things appropriately.

Regardless, be sure to call out things that you don't agree with (in comments) and/or you have better suggestions. I'm also not always right and my suggestions might not align with what the code represents etc., so make sure to call these out.

azure-pipelines.yml
```diff
@@ -0,0 +1,191 @@
...
+          displayName: 'Flutter build - iOS'
+          inputs:
+            target: 'ios'
+            projectDirectory: 'device_calendar/example'
+            iosCodesign: false
+            
+  # -----------CD Pre-release -----------
+  - stage: DevelopmentRelease
+    dependsOn: Build
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/release'), eq(variables['Build.SourceBranch'], 'refs/hotfix'))
```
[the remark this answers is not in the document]
Brett09: Found a way to do this (and tested): using `in()` condition: `Evaluates True if left parameter is equal to any right parameter`

fyi, we were using wrong variables for branches: it supposed to be something like `refs/heads/release`. It didn't run at all when I used `refs/release`

azure-pipelines.yml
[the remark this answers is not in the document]
Brett09: Didn't really find any other strings that should be used as a global variable :\

azure-pipelines.yml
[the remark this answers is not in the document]
Brett09: Added a task: https://dev.azure.com/builttoroam/Flutter%20Plugins/_workitems/edit/20527

azure-pipelines.yml
Brett09: Added a task: https://dev.azure.com/builttoroam/Flutter%20Plugins/_workitems/edit/20528

azure-pipelines.yml
Brett09: Nay, that was the only line it uses `versionRegex` for a replace, so I'll leave it as it is

azure-pipelines.yml
```diff
@@ -0,0 +1,189 @@
...
+          displayName: 'Flutter build - iOS'
+          inputs:
+            target: 'ios'
+            projectDirectory: 'device_calendar/example'
+            iosCodesign: false
+            
+  # -----------CD Pre-release -----------
+  - stage: DevelopmentRelease
+    dependsOn: Build
+    condition: and(succeeded(), in(variables['Build.SourceBranch'], 'refs/heads/release', 'refs/heads/hotfix'))
```
Brett09: @mkieres @nickrandolph just came up with a thought that `Build.SourceBranch` isn't right. I believe it should be `Build.TargetBranch`.

When I want to pre-release a version, we're creating a PR from `develop` branch and merging into `release`. In this case, the source branch is `develop` and target branch is `release`.
If we use `Build.SourceBranch`, then the CD Pre-release will never run because the condition is looking for a source branch of `release` or `hotfix` where it's passing `develop`.

What you're thoughts?

Brett09: But little worried here if we use `Build.TargetBranch`, it might trigger the CD when I create a new PR or update something in a PR?

azure-pipelines.yml
```diff
@@ -0,0 +1,189 @@
+# Starter pipeline
```
nickrandolph: Remove these header comments

azure-pipelines.yml
nickrandolph: I think SourceBranch is ok for the moment because it just means that this will trigger once the PR has been merged. We might eventually look at making this a pre-merge action but for now I think this is ok

azure-pipelines.yml
```diff
@@ -0,0 +1,189 @@
...
+        vmImage: 'windows-latest'
+       steps:
+        - task: FlutterInstall@0
+          displayName: 'Flutter install'
+          inputs:
+            channel: 'stable'
+            version: 'latest'
+
+        - task: PowerShell@2
+          condition: eq(variables['Build.SourceBranch'], 'refs/heads/release')
```
nickrandolph: Do we need this condition here - surely we want to do this for hotfix as well as release branches?

azure-pipelines.yml
```diff
@@ -0,0 +1,189 @@
...
+          inputs:
+            githubEndpoint: 'GitHub Tagging'
+            manuallySetRepository: false
+            githubRepository: 'builttoroam/flutter_plugins'
+            githubTag: 'v$(currentVersion)'
+            githubReleaseTitle: '$(versionSummary)'
+            githubReleaseDraft: false
+            githubReleasePrerelease: false
+            githubIgnoreAssets: false
+            githubReleaseAsset: '$(Build.ArtifactStagingDirectory)/*'
```
nickrandolph: Does setting this push extra content to the release?

azure-pipelines.yml
```diff
@@ -0,0 +1,189 @@
...
+            githubReuseRelease: false
+            githubReuseDraftOnly: false
+            githubSkipDuplicatedAssets: false
+            githubEditRelease: false
+            githubDeleteEmptyTag: false
+
+        - task: CmdLine@2
```
nickrandolph: Wouldn't we want to do the dry run publish before we do the github tag?

Brett09: I think someone pointed out about this but can't remember, happy to move dry run & publish prior to github tagging

azure-pipelines.yml
Brett09: You mean the source code files? I haven't tested the actual tag (only tested drafts), but `'$(Build.ArtifactStagingDirectory)/*'` this block was provided automatically provided from the default github tagging task

azure-pipelines.yml
Brett09: That condition (`condition: eq(variables['Build.SourceBranch'], 'refs/heads/release')`) is only true when the branch is `release` isn't it? Yes we only want to run the task if the branch is `release`, not `hotfix`

azure-pipelines.yml
Brett09: Actually I reckon this could be a better solution as the github tag will not be created if publish fails - which means I don't need the task with the `failed()` condition to delete the github tag

azure-pipelines.yml
Brett09: Updated this to publish dev versions for `hotfix` branch as well

nickrandolph: As discussed, remove this condition

azure-pipelines.yml
```diff
@@ -0,0 +1,174 @@
...
+              [bool] $hasMatches = $pubspecContent -match $versionRegex
+              
+              [string] $newPubspecContent = $pubspecContent -replace $version, "$version-dev.$(Build.BuildID)"
+
+              Write-Host " (i) Updated pubspec.yaml content: $newPubspecContent"
+
+              $newPubspecContent | Set-Content -Path "$(Build.SourcesDirectory)/device_calendar/pubspec.yaml"
+
+        - task: PowerShell@2
+          condition: eq(variables['Build.SourceBranch'], 'refs/heads/release')
```
nickrandolph: and remove this condition

azure-pipelines.yml
nickrandolph: k leave as is, and we'll see what it generates

azure-pipelines.yml
Brett09: Seems like you're looking at the old commit, I've removed them already

PR #178, opened 2020-03-06
Dual screen

NickBalnaves:


dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+    ActivityAware {
+    private val METHOD_CHANNEL_NAME = "plugins.builttoroam.com/dual_screen/methods"
+    private val EVENT_CHANNEL_NAME = "plugins.builttoroam.com/dual_screen/events"
+    private lateinit var methodChannel: MethodChannel
+    private lateinit var eventChannel: EventChannel
+    private var context: Context? = null
+    private var activity: Activity? = null
+
+    fun registerPlugin(context: Context?, messenger: BinaryMessenger?) {
+        activity = this.activity
```
mkieres: I'm a bit puzzled by this code:

1. What is being assigned to what? Is the this.activity the class activity and the activity is just a local variable?
2. Isn't the context the activity anyways?

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+    override fun onAttachedToActivity(binding: ActivityPluginBinding) {
+        this.activity = binding.activity
+    }
+
+    override fun onReattachedToActivityForConfigChanges(binding: ActivityPluginBinding) {
+        this.activity = binding.activity
+    }
+
+    override fun onDetachedFromActivityForConfigChanges() {
+        this.activity = null
```
mkieres: I'm not sure of the purpose of these two calls that are nullifying the activity but should we do similar for context? That is If this is part of the disposing process.

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    override fun onDetachedFromActivity() {
+        this.activity = null
+    }
+
+    fun isDualScreenDevice(): Boolean {
+        return context?.getPackageManager()!!.hasSystemFeature(
```
mkieres: Isn't this going to crash the app if `getPackageManger` will be null?

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+            "com.microsoft.device.display.displaymask"
```
mkieres: This should be a extracted as a const

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+        )
+    }
+
+    fun isAppSpanned(): Boolean {
+        if (isDualScreenDevice()) {
```
mkieres: I would probably do the same thing as mentioned earlier, with trying to return as soon as possible and avoid nesting

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+            var boundings = DisplayMask.fromResourcesRectApproximation(this.activity).getBoundingRects()
+            if (boundings.isEmpty()) {
+                return false
+            }
+            var drawingRect = android.graphics.Rect()
+            activity?.getWindow()?.getDecorView()?.getRootView()?.getDrawingRect(drawingRect)
```
mkieres: I'm puzzled by this activity vs this.activity again. Maybe I just don't understand how dart plugins work but do we need both and how do they differ?

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+            return boundings.get(0).intersect(drawingRect)
```
mkieres: Why do we use `0` here? We should create a variable that will describe its purpose or make a comment. I would suggest to go with a variable / const approach.

NickBalnaves: I've added a comment, i think the const approach is a bit unnecessary

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        when (call.method) {
+            "isDualScreenDevice" -> {
+                result.success(isDualScreenDevice())
+            }
+            "isAppSpanned" -> {
+                if (isDualScreenDevice()) {
```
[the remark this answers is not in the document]
NickBalnaves: I would usually tend to agree with this approach if we were returning a value for each possibility, however since onMethodCall is a void function i dont think we should be returning a value here

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
mkieres: I disagree. Random variables used in code (like this one) don't have any meaning and are a pain to decypher when reading and understanding code. We should always try to make the code as much readable as possible.

It might be that to make this code more readable we should use different approach and assign the `boundings.get(0)` to a variable with a name that would describe the boundries element, instead of extracing 0 to a const.

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+    companion object {
+        @JvmStatic
+        fun registerWith(registrar: Registrar) {
+            DualScreenPlugin().registerPlugin(registrar.context(), registrar.messenger())
+        }
+    }
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        when (call.method) {
+            "isDualScreenDevice" -> {
```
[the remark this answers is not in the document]
NickBalnaves: Extracted out

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+                oldTop: Int,
+                oldRight: Int,
+                oldBottom: Int
+            ) {
+                events!!.success(isAppSpanned())
+            }
+        })
+    }
+
+    override fun onCancel(arguments: Any?) {
```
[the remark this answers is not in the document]
NickBalnaves: Yes its a required override method since we are using the ActivityAware interface

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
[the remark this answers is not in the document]
mkieres: Oh, my bad. You're right. I don't know why I assumed we're returning stuff here.

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: The activity initialisation wasn't needed here so i have removed that line

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: @nickrandolph which approach would you like to take?

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: My thoughts for only including activity here were that this method is only related to activity changes and not context changes

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: Extracted

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: Yep you're right, I've added in a check

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
```diff
@@ -0,0 +1,132 @@
...
+                left: Int,
+                top: Int,
+                right: Int,
+                bottom: Int,
+                oldLeft: Int,
+                oldTop: Int,
+                oldRight: Int,
+                oldBottom: Int
+            ) {
+                events!!.success(isAppSpanned())
```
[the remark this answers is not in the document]
NickBalnaves: Yep, I've added in a check

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: Using 'this' is just explicit that we want the activity defined in the class. Usually helpful when you have a variable named the same thing that you want to assign to. Originally these methods had an activity parameter but has been changed to binding.activity so ive removed 'this' references

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
NickBalnaves: I've removed the else

dual_screen/android/src/main/kotlin/com/builttoroam/dual_screen/DualScreenPlugin.kt
nickrandolph: Is there not a "first()" method that we can use instead of get(index)

NickBalnaves: There is! pushed a changed to use first()

Brett09:
Project moved to [here](https://github.com/builttoroam/dual_screen), thus closing this pr

Issue #182, opened 2020-03-11
Incorrect value set to Events.EVENT_TIMEZONE (Android)


kinex:
I have been investigating some timezone related issues and found this in the plugin code:
https://github.com/builttoroam/flutter_plugins/blob/develop/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt#L390

I am quite sure `displayName` is not the correct value to put there, you should use `ID` instead (as told e.g. [here](https://stackoverflow.com/questions/11934465/eventtimezone-value)).

So why does it seem to work then without issues? I think this is because the value (`DateTime.millisecondsSinceEpoch`) passed from Dart is UTC time. Setting invalid value to `EVENT_TIMEZONE` is probably ignored and actual timezone is reverted to UTC. 

So I think it is also a bug even trying to put the current device timezone to EVENT_TIMEZONE, because the `Event.start` is UTC. 

Fix options:
1. Save in `EVENT_TIMEZONE` the UTC timezone.
2. Convert Event start/end to local time and save in `EVENT_TIMEZONE` the device timezone.
3. Remove setting `EVENT_TIMEZONE` at all? Maybe the default is UTC? Should be tested.

Or have I misunderstood something?

Brett09:
I haven't seen any issue with using `displayName` except it uses wrong (or null) timezone when you use your local calendar in Samsung phones. Just tried out with using `id` and it seems to be working.
Thanks for the point and I'll put this in so it's available in the next release.

Brett09:
All `TimeZone` variables are now using `id` in v3.1.0, closing this ticket.

Issue #164, opened 2020-02-03
Allday property resulting in incorrect event 


nickrandolph:
Reported issue: 'the property "Event.allDay = true;" doesn't save the event correctly in the calendar.'

Further comments: this appears to be an issue on Android based on the timezone of the event (eg setting an all day event with a start day before 11am AEST results in the event being for the previous day)

Brett09:
Adding info from Android Developer's page:

- If allDay is set to 1 eventTimezone must be Time#TIMEZONE_UTC and the time must correspond to a midnight boundary.

https://developer.android.com/reference/android/provider/CalendarContract.Events#writing-to-events

Bernhard7:
But the class "Event" in "device_calendar" has no property "timezone".

Brett09:
@Bernhard7, this issue is only for Android so I'll need to do a fix on Android native code, not Flutter.

Bernhard7:
Thanks for updating device_calendar to ^3.0.0+3. Now it works.

kinex:
I think `retrieveEvents` still returns events with invalid start/end times (testing in master). Few related observations: https://github.com/builttoroam/flutter_plugins/blob/develop/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt#L511
=> Should you handle EVENT_TIMEZONE and EVENT_END_TIMEZONE in that code? https://developer.android.com/reference/android/provider/CalendarContract.EventsColumns#EVENT_TIMEZONE

I was also wondering if this code works as expected:
https://github.com/builttoroam/flutter_plugins/blob/develop/device_calendar/lib/src/models/event.dart#L64
Have you noticed that `DateTime.fromMillisecondsSinceEpoch` has a parameter `isUtc` which is `false` by default. I would bet it should be set to `true`, assuming the dates returned from plugin are meant to be UTC time.

Brett09:
@kinex I've already found a solution for this and will be fixed in next release

kinex:
@Brett09 Ok, great. And hopefully you noticed that my previous comment is probably related to a different issue than originally reported in this issue (sorry my bad, I should have created a new issue). In my opinion this is the most annoying issue at the moment what I have seen in this plugin, as for example all events retrieved from holiday calendars have invalid timestamps. Waiting for the fixes, thanks!

Brett09:
This has been fixed in v3.1.0. You will be able to set all day events without any time issues in Android

[...]

PR #183, opened 2020-03-16
Allday timezone fix (Android)

Brett09:
Issue: #164 
Also updated to use `id` property instead of `displayName` as from: #182

### DIFF
--- a/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
+++ b/device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
@@ -12,6 +12,7 @@ import android.database.Cursor
 import android.net.Uri
 import android.provider.CalendarContract
 import android.provider.CalendarContract.Events
+import android.text.format.DateUtils
 import io.flutter.plugin.common.PluginRegistry.Registrar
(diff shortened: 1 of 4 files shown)


device_calendar/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -371,12 +372,31 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
...
+            val utcTimeZone =  TimeZone.getTimeZone("UTC")
+
+            val calendar = java.util.Calendar.getInstance(utcTimeZone)
+            calendar.timeInMillis = event.start!!
+            calendar.set(java.util.Calendar.HOUR, 0)
+            calendar.set(java.util.Calendar.MINUTE, 0)
+            calendar.set(java.util.Calendar.SECOND, 0)
+            calendar.set(java.util.Calendar.MILLISECOND, 0)
+
+            val allDayTime = calendar.timeInMillis + DateUtils.DAY_IN_MILLIS
```
nickrandolph: Not sure this is required.

Brett09: The logic was a little wrong (I had to assign the timezone for `calendar` object after assigning the time), updated and tested successfully

Issue #179, opened 2020-03-06
Add ACCOUNT_NAME and ACCOUNT_TYPE to Calendar data


kinex:
Calendar data is missing ACCOUNT_NAME and ACCOUNT_TYPE (referring to Android API terms). Those fields would allow displaying prettier list of available calendars to users, for example grouped by account name.

https://developer.android.com/reference/android/provider/CalendarContract.SyncColumns#ACCOUNT_NAME

Brett09:
Do you know if there are iOS equivalent variables? We need to have the consistency between platforms.

Brett09:
I've found iOS variables and applied them. I'll try to add this to the next release.

fyi, iOS uses `EKSource` for account information [found here](https://developer.apple.com/documentation/eventkit/eksource)

Brett09:
v3.1.0 has been published and you'll be able to get account name and type under `Calendar` model

PR #189, opened 2020-03-19
Add Calendar & Retrieve Account Name/Type

Brett09:
Issues:
- Calendar: #115 
- Account name/type: #179 

Notes:
A colour for the calendar can be chosen when creating a new calendar


device_calendar/example/lib/presentation/pages/calendar_add.dart
```diff
@@ -0,0 +1,149 @@
...
+                    items: ColorChoice.values
+                        .map((color) => DropdownMenuItem(
+                              value: color,
+                              child: Text(color.toString().split('.').last),
+                            ))
+                        .toList(),
+                  ),
+                ],
+              ),
+              if (Platform.isAndroid) ...[
```
nickrandolph: I don't think the ...[ is required here is it?

device_calendar/ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -688,5 +753,31 @@ extension UIColor {
...
+                var hexNumber: UInt64 = 0
+
+                if scanner.scanHexInt64(&hexNumber) {
+                    a = CGFloat((hexNumber & 0xff000000) >> 24) / 255
+                    r = CGFloat((hexNumber & 0x00ff0000) >> 16) / 255
+                    g = CGFloat((hexNumber & 0x0000ff00) >> 8) / 255
+                    b = CGFloat((hexNumber & 0x000000ff)) / 255
+                    
+                    self.init(red: r, green: g, blue: b, alpha: a)
+                    return
```
nickrandolph: should this return a value?  a couple of lines below it returns nil, so seems weird that this return doesn't actually return anything

device_calendar/example/lib/presentation/pages/calendar_add.dart
Brett09: argh I need to get a habit of not using the array block `...[]` for a single widget. I'll remove this in the next PR

device_calendar/ios/Classes/SwiftDeviceCalendarPlugin.swift
Brett09: Yeah tested that it returns a wrong colour if I remove that `return` and build fails if I remove `return nil` on the few line below

Issue #181, opened 2020-03-09
Use threads for better performance


kinex:
I noticed this plugin is doing its work in UI thread. I can already see some slowness in my app UI. Could you please consider applying threading for better performance. This is also recommended at least in related Android [documentation](https://developer.android.com/reference/android/provider/CalendarContract.Instances#query(android.content.ContentResolver,%20java.lang.String%5B%5D,%20long,%20long)) .

Threading is very simple in Kotlin and in Swift. If you need help, here are some code templates (this is how I do it in my own plugins):

Kotlin:
```kotlin
val uiScope = CoroutineScope(Dispatchers.Main)
uiScope.launch {
    // notice placement of try/catch
    try {
        withContext(Dispatchers.IO) {
            // do heavy work here
        }
        // this must be called on UI thread
        result.success(true)
    } catch (e: Exception) {
        // this must be called on UI thread
        result.error("error", e.localizedMessage, e.toString())
    }
}
```

Swift:
```swift
// you may also consider other options than .userInitiated
DispatchQueue.global(qos: .userInitiated).async {
    do {
        // do heavy work here
        
        DispatchQueue.main.async {
            // this must be called on UI thread
            result(true)
        }
    } catch {
        DispatchQueue.main.async {
            // this must be called on UI thread
            result(FlutterError(code: "ERROR",
                                message: error.localizedDescription,
                                details: nil))
        }
    }
}
```

Brett09:
Thanks for the suggestion & examples. Will look into them and apply them when I get a chance.

intkhabahmed:
@Brett09 what is the progress in this, it is causing my app to skip frames, please resolve this asap.

dungtang93:
> @Brett09 what is the progress in this, it is causing my app to skip frames, please resolve this asap.

https://github.com/builttoroam/device_calendar/pull/222#issuecomment-643838348

Android only, as I don't have an iOS phone to code & test this out, help wanted

imhafeez:
`createOrUpdateEvent` is also on main thread. it causes frame skips as well. Please change it to background.

dungtang93:
@imhafeez Why are you experiencing frame skip in `createOrUpdateEvent`? are you creating a lot of events at once?

imhafeez:
@dungtang93 yes. actually I was syncing all the events from server to my local calendar. any solution for that?

thomassth:
Seems that we have threading on Android but not on iOS.

[...]

PR #222, opened 2020-05-06
Enable threading (Android) for better performance

dungtang93:
#181 

* Currently applied to Android only
* On Android devices, `retrieveEvents()` use thread by default

### DIFF
--- a/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
+++ b/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
@@ -67,6 +67,7 @@ import com.google.gson.GsonBuilder
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.PluginRegistry
 import io.flutter.plugin.common.PluginRegistry.Registrar
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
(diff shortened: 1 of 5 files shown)


nickrandolph:
@dungtang93 is there a reason why we need a separate method for this? If we can make the existing method non-blocking that would be perferable.

dungtang93:
Yes you're right, non-blocking is preferable in every situations, I'll change the current methods instead of making new ones soon.

nickrandolph:
Thanks @dungtang93 - let me know when you've pushed the changes and I'll review.

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -259,7 +262,6 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
         finishWithSuccess(calendarId.toString(), pendingChannelResult)
     }
 
-    @SuppressLint("MissingPermission")
```
nickrandolph, line 262: Is this suppression no longer required?

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -281,40 +283,40 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
             val eventsUri = eventsUriBuilder.build()
             val eventsCalendarQuery = "(${Events.CALENDAR_ID} = $calendarId)"
             val eventsNotDeletedQuery = "(${Events.DELETED} != 1)"
-            val eventsIdsQueryElements = eventIds.map { "(${CalendarContract.Instances.EVENT_ID} = $it)" }
-            val eventsIdsQuery = eventsIdsQueryElements.joinToString(" OR ")
+            val eventsIdsQuery ="(${CalendarContract.Instances.EVENT_ID} IN (${eventIds.joinToString()}))"
```
nickrandolph, line 287: Is this a fix, or just tidying up the code?

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -281,40 +283,40 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
...
 
             var eventsSelectionQuery = "$eventsCalendarQuery AND $eventsNotDeletedQuery"
-            if (eventsIdsQuery.isNotEmpty()) {
+            if (eventIds.isNotEmpty()) {
                 eventsSelectionQuery += " AND ($eventsIdsQuery)"
             }
-            val eventsSortOrder = Events.DTSTART + " ASC"
+            val eventsSortOrder = Events.DTSTART + " DESC"
```
nickrandolph, line 292: Was switching the order of events intentional?

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -281,40 +283,40 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
...
+                }.invokeOnCompletion {
+                    eventsCursor?.close()
+                    _registrar!!.activity().runOnUiThread {
+                        finishWithSuccess(_gson?.toJson(events), pendingChannelResult)
+                    }
                 }
             } catch (e: Exception) {
                 finishWithError(GENERIC_ERROR, e.message, pendingChannelResult)
-            } finally {
-                eventsCursor?.close()
```
nickrandolph, line 314: If exceptions happen within the globalscope, does the exception get propagated out to the wrapper try-catch? and are we able to force the eventsCursor to be closed?

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
dungtang93: Yes I don't see any warnings in the method after removing it

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
dungtang93: *Not* a fix or a performance improvement, just tidy up the syntax a bit

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
dungtang93: A mistake on my part, I'll change it back

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
nickrandolph: I've raised a new issue to track adding better error handling

nickrandolph:
@dungtang93  There is a dev release available - please it out and check for any issues - https://pub.dev/packages/device_calendar/versions/3.2.0-dev.11602

Issue #223, opened 2020-05-12
iOS createCalendar error: Local calendar was not found


thiszhong:
It occurs on my real iPhone device(iPhone 7, iOS 13.3.1 et.), while simulator and android are OK. I tried both my app and the demo.
```
[599] Device calendar plugin ran into an issue. Platform specific exception [500], with message: "Local calendar was not found.", has been thrown.
```
![WechatIMG1](https://user-images.githubusercontent.com/12380965/81690907-92f4b900-948e-11ea-8562-ebcff2331769.png)

```
flutter doctor -v
[✓] Flutter (Channel stable, v1.17.0, on Mac OS X 10.15.4 19E287, locale zh-Hans-CN)
    • Flutter version 1.17.0 at /Users/zhong/code/flutter
    • Framework revision e6b34c2b5c (10 days ago), 2020-05-02 11:39:18 -0700
    • Engine revision 540786dd51
    • Dart version 2.8.1

 
[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /Users/zhong/Library/Android/sdk
    • Platform android-29, build-tools 28.0.3
    • ANDROID_HOME = /Users/zhong/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 11.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.4.1, Build version 11E503a
    • CocoaPods version 1.9.0

[✓] Android Studio (version 3.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 31.3.1
    • Dart plugin version 181.5656
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)

[✓] VS Code (version 1.44.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.10.1

[!] Connected device
    ! No devices available
```

thiszhong:
Seems to be the iCloud. When I close it by setting-account-iCloud-Calendar, createCalendar successed. Is there any solution?

zeshanghayoomi:
are you guys gonna release the fix for this anytime soon ?

[...]

PR #224, opened 2020-05-12
[WIP] fix createCalendar on iOS device when iCloud is open

thiszhong:
#223 
This resolved my problem. But I am not a iOS developer, So have a check, please.

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -161,7 +161,7 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin {
                 calendar.cgColor = UIColor(red: 255, green: 0, blue: 0, alpha: 0).cgColor // Red colour as a default
             }
             
-            let localSources = eventStore.sources.filter { $0.sourceType == .local }
+            let localSources = eventStore.sources.filter { $0.sourceType == .local || ($0.sourceType == .calDAV && $0.title == "iCloud") }
             
             if (!localSources.isEmpty) {
                 calendar.source = localSources.first


nickrandolph:
@thiszhong can you confirm that the iCloud calendar always has the title "iCloud" (perhaps links to any docs). 
Am I correct in saying that when user has iCloud enabled, they don't see any local source types?

thiszhong:
> @thiszhong can you confirm that the iCloud calendar always has the title "iCloud" (perhaps links to any docs).
> Am I correct in saying that when user has iCloud enabled, they don't see any local source types?

Actually I'm not sure because I don't know much about ios. But it's the best answer I can find.
[https://stackoverflow.com/a/15980556](https://stackoverflow.com/a/15980556) 

[https://github.com/TheAlmightyBob/Calendars/issues/22#issuecomment-236888276](https://github.com/TheAlmightyBob/Calendars/issues/22#issuecomment-236888276) 

[https://oleb.net/blog/2012/05/creating-and-deleting-calendars-in-ios/](https://oleb.net/blog/2012/05/creating-and-deleting-calendars-in-ios/) 

> Update February 12, 2014 and March 05, 2014: Robert de Jong pointed out to me that the source title is indeed user-editable (Settings > iCloud > Account > Description), which makes this way of detecting the iCloud calendar store unreliable. Robert suggests this approach:

`EKSource *theSource = [[eventStore defaultCalendarForNewEvents] source];`

> For most users, this should provide the iCloud source, if it’s set up, or the local source, if not. But it might as well return a totally different source if the user has configured another default. There seems to be no 100% reliable way to retrieve the iCloud source with public API.  

[Apple Watch for Developers - p44](https://books.google.com/books?id=RL9OCgAAQBAJ&pg=PA44&lpg=PA44&dq=ios+calendar+sourceType.title+icloud&source=bl&ots=FC5NlKRLYT&sig=ACfU3U2tqZK0D1ucKNO2TPGbpJQ3wdLTQg&hl=zh-CN&sa=X&ved=2ahUKEwj5-KWemfHpAhVHtZ4KHZhbB4QQ6AEwAXoECAoQAQ#v=onepage&q=icloud&f=false)



Feel free to close it if there is a better answer.

emm... I think it is a check when local source is nil.

nickrandolph:
@thiszhong I think this should be a fall back process - look for .local source type. If that doesn't exist, then use
EKSource *theSource = [[eventStore defaultCalendarForNewEvents] source];
to find the appropriate source.
I'm not convinced that using the source of the defaultCalendarForNewEvents is always the correct behaviour. Hence using it only as a fallback. 
Thoughts?

petekblaz:
Any info on when this will be merged?

nickrandolph:
@petekblaz see my comment above - I think changes are required before we can merge this.

hnvn:
Why don't use `eventStore.defaultCalendarForNewEvents` ?

hnvn:
Why don't use `eventStore.defaultCalendarForNewEvents.source`. I would like to handle it like:
```swift
guard let defaultSource = eventStore.defaultCalendarForNewEvents?.source else {
   result(FlutterError(code: self.genericError, message: "Default calendar was not found.", details: nil))
   return
}
                
calendar.source = defaultSource;
                
try eventStore.saveCalendar(calendar, commit: true)
result(calendar.calendarIdentifier)

```

rjgpereira:
Hi guys! 

I think it would be useful to give a chance to pick the calendar source. So, my suggestion is creating a methodCall to retrieve the source and pass it to the "createCalendar". Otherwise, you would default to the defaultCalendarForNewEvents like @hnvn is suggesting.

What do you think?

thiszhong:
A new approach seems to be needed.

elaishane:
Hey guys is this issue fixed/not. My clients are asking for a solution. Somebody please give a solution asap

michaelchiowong:
hi, guys, any updates for this ?

sj1914:
Would love a solution to this, i think the fallback option sounds like a good alternative to the current behaviour!

At the moment the only option, would love to know if someone has thought of an alternative or workaround, seems to be to provide some troubleshooting to users to tell them to turn off icloud while we sync their calendar for the first time, which doesn't seem ideal.

I want to keep events created by my app separate as it makes syncing and unsyncing more manageable. Without the ability to create a new calendar on iOS I'm super stuck. Help!

PR #231, opened 2020-05-28
Fixed two links

rmhughes:
- Fixed the links for calendar_event and calendar_events


nickrandolph:
These links will be correct once the changes in develop are promoted to master.

PR #232, opened 2020-05-28
Bug fix when adding new calendar.

Pranjal2041:
Fixed bug giving error when user tried to add a calendar while gmail is open.


elaishane:
Dear Admin this pull request is important. Because without this implementation the calendar won't work on Samsung Phones. I had to implement it the hard way by going into the plugin and modifying it. If possible please try to merge it as soon as possible because without this pull request the calendar add won't work on android Samsung phones

PaulBout1:
Please merge this in a released branch. current version `3.1.0` does not implement it

Issue #235, opened 2020-06-14
Change Availability from String to Enum


nickrandolph:
PR #230 introduces an availability property. 

Suggestion is to change the data type from string to an enum with values free, busy, tentative, unavailable (null for any other state).

carlosDigio:
Hi @nickrandolph . We have implemented this issue in PR #238. Can you check it out?

PR #238, opened 2020-06-16
Changing Availability to Enum

carlosDigio:
Hi @nickrandolph . We have implemented the issue #235 in this PR. Can you check it out?
Thanks


android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
```diff
@@ -1,20 +1,20 @@
...
 import android.content.pm.PackageManager
 import android.database.Cursor
+import android.graphics.Color
 import android.net.Uri
 import android.provider.CalendarContract
 import android.provider.CalendarContract.CALLER_IS_SYNCADAPTER
 import android.provider.CalendarContract.Events
 import android.text.format.DateUtils
-import com.builttoroam.devicecalendar.common.Constants
+import com.builttoroam.devicecalendar.common.*
```
nickrandolph, line 17: Will this import all the nested constants as well eg Constants.Companion.ATTENDEE_EMAIL_INDEX.
If so, can we remove the individual imports below

android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
```diff
@@ -223,4 +220,11 @@ class DeviceCalendarPlugin() : MethodCallHandler {
     private inline fun <reified T : Any> Any?.toMutableListOf(): MutableList<T>? {
         return this?.toListOf<T>()?.toMutableList()
     }
+
+    private fun parseAvailability(value: String?): Availability? =
+            if (value == null || value == "UNAVAILABLE") {
```
nickrandolph, line 225: Please extract "UNAVAILABLE" string literal out as a constant

example/lib/presentation/pages/calendar_event.dart
```diff
@@ -843,6 +849,21 @@ class _CalendarEventPageState extends State<CalendarEventPage> {
     }
   }
 
+  Text _availabilityToText(Availability availability) {
+    switch (availability) {
```
nickrandolph, line 853: I'd suggest creating a const map from Availability to string and then just doing a lookup instead of this switch

android/src/main/kotlin/com/builttoroam/devicecalendar/DeviceCalendarPlugin.kt
alvaroDigio: Fixed

example/lib/presentation/pages/calendar_event.dart
alvaroDigio: Fixed

android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
alvaroDigio: Fixed

carlosDigio:
> I'm not sure we need to define Availability enum in all three languages (ie dart, kt, swift) since we need to serialise it via the channel anyhow. Wouldn't it be easier to keep the use of string for availability in kt/swift and then just decode to an enum in dart?

Hi! It may not be entirely necessary in this case, but we think that, just as in dart seems correct, in native it would also be more appropriate to use enums rather than strings despite having to do a decoding for use with dart or any other language.

nickrandolph:
@carlosDigio for some reason the Android build is failing - can you see if you can resolve this please

alvaroDigio:
@nickrandolph fixed the build error

Issue #242, opened 2020-06-18
iOS daysOfWeek 


blehr:
On iOS when added the daysOfWeek List to the recurrenceRule, only the first day in the list is added when an event is created.

nickrandolph:
Can you add this case to the example app and submit as a wip PR so we can diagnose what's going on

PR #243, opened 2020-06-18
iOS daysOfWeek

blehr:
#242

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -455,9 +455,7 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin {
                             weekNumber: weekOfMonth
                         ))
                     }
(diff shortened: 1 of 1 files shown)


ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -363,7 +363,7 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin {
             
             var totalOccurrences: Int?
             var endDate: Int64?
-            if(ekRecurrenceRule.recurrenceEnd?.occurrenceCount != nil) {
+            if(ekRecurrenceRule.recurrenceEnd?.occurrenceCount != nil  && ekRecurrenceRule.recurrenceEnd?.occurrenceCount != 0) {
```
nickrandolph, line 366: Is this change required?  Is there a scenario where occurrenceCount==0 and setting totalOccurrences isn't correct?

PR #252, opened 2020-07-01
Fix bug not found local calendar

hnvn:


nickrandolph:
@hnvn thanks for the contribution.
My only concern with this (as I pointed out in https://github.com/builttoroam/device_calendar/pull/224) is that this assumes that the source for the "DefaultCalendarForEvents" should be use for new calendars. I don't think this is the intended behaviour of this setting on iOS and think that it would be better to default to a new local calendar.

hnvn:
You assume that better default calendar is local calendar, but the local source type won't be there in case of iCloud on, so we end up with the feature not working and the plugin also don't provide any alternative configurable argument to make it working.

nickrandolph:
Hence my suggestion as part of #224, if the local source type exists it should be the first choice, with a fall back to the source of the default calendar.
I suspect that long term we may need to provide the ability for the developer to specify what source to use.

PR #297, opened 2021-04-26
Timezone implementation

andzejsw:
Android implementation should now work with Timezone logic. So should example build. 
Currently haven't looked on iOS.

PR #299, opened 2021-05-05
Improve pub points by 10

thomassth:
No actual functional difference. Only some formatting.

Apparently this makes the code "has no errors, warnings, lints, or formatting issues" so  ¯\\_(ツ)_/¯

- [ ] Squash commit


andzejsw:
How is this PR different from those commits merged in latest project commits here: https://github.com/builttoroam/device_calendar/commits/develop

andzejsw (dismissed):

nickrandolph:
Is this PR required now?

thomassth:
I think #297 should do the same thing, as flutter plugins in code editors should auto format when saving files.

Otherwise I'll just do it after that merged.

PR #300, opened 2021-05-07
Calendar create issue on Android Samsung device

imhafeez:


andzejsw:
This PR contains only formatting change

imhafeez:
@andzejsw There are two commits. changes are in the first commit. and the second I mistakenly import the color two times then fix it.

PR #304, opened 2021-05-11
iOS fixes

thomassth:
#302 but in a separate branch. See if this is better.


nickrandolph:
I think since we merged the timezone pr that there's now conflicts on this PR that are preventing a merge:
Conflicting files
example/ios/.symlinks/plugins/device_calendar
example/lib/presentation/pages/calendars.dart

example/ios/.symlinks/plugins/device_calendar
```diff
@@ -1 +1 @@
-/Volumes/MacData/Codebase/Flutter/device_calendar/
```
nickrandolph, line 1: @thomassth is this change in folder name required?

andzejsw:
In my opinion no generated files should be commited and should be added to ".gitignore". Since "ios/.symlinks" folder is generated, it with all of its contend should be gitignored. Do you agree guys?

thomassth:
i've added it into .gitignore and caught up with upstream.

nickrandolph:
@thomassth these changes look fine but can we link to the relevant issue here - I'm not sure what exactly we're fixing

thomassth:
I was looking at [this comment](https://github.com/flutter/flutter/issues/16049#issuecomment-611192738).

The whole thread was repeatedly referenced in out issues and readme.

Not sure what it fixes, as I didn't see any error on my side before, but it seems needed.

nickrandolph:
@andzejsw  and @thomassth are you ok with this PR?

andzejsw:
All looks good to me

sowens-csd (approved):
Changes look safe to me. Ran successfully on both an iOS 15 simulator and a real device.

PR #311, opened 2021-06-15
iOS Source fallbacks and re-add url parameter

rmtmckenzie:
I should have separated these into two pull requests but I believe they are both worthwhile fixes; if you disagree about the source fallbacks then it would be great if you could at least re-add the url parameter.

In production I'm seeing errors and a failure to create scheduled notifications on many devices, as it seems in more recent iOS versions there is often no local calendar. It would be even better if there were a way to query sources directly from flutter, but that would complicate things and is probably overkill in most cases - I'm going to guess that 99% of users just use the default iCloud calendar.

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -159,6 +159,26 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin {
         result(hasPermissions)
     }
     
+    private func getSource() -> EKSource? {
+      let localSources = eventStore.sources.filter { $0.sourceType == .local }
+
+      if (!localSources.isEmpty) {
+        return localSources.first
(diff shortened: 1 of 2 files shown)


nickrandolph:
@andzejsw  and @thomassth are you both ok with this PR?

andzejsw:
@nickrandolph All looks good.

saschaernst:
Will this PR ever be merged?

rmtmckenzie:
lmk know if you need anything else from my end to make it happen!

thomassth (approved):

thomassth:
I don't have an iOS device at hand

Can someone run this PR on device and see if there's any error popping up

In case there is some migration issue from previous versions

rmtmckenzie:
If it helps, I've been running this in production with roughly 30k active iOS users for 3 months and haven't been seeing any errors.

PR #314, opened 2021-07-05
Make timezone database initialization optional

kightsonsanom:
1. In case the application initializes its own timezone by setting the local timezone as the current device timezone, it should be possible to omit the initialization of the timezone database so that it doesn't default to UTC.
2. Update `flutter_native_timezone` library to the latest 2.0.0 version.

### DIFF
--- a/lib/src/device_calendar.dart
+++ b/lib/src/device_calendar.dart
@@ -24,8 +24,10 @@ class DeviceCalendarPlugin {
 
   static final DeviceCalendarPlugin _instance = DeviceCalendarPlugin.private();
 
-  factory DeviceCalendarPlugin() {
-    tz.initializeTimeZones();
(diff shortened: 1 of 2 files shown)


thomassth:
Nice, didn't think it would be easy to implement.

How does it behave when upgrading from 3.x or the current 4.x prereleases?

ie. how many lines would people need to fix when upgrading

kightsonsanom:
No changes will be required, this is an optional parameter with a default value so the call to `DeviceCalendarPlugin()` will always initialize the timezone database as  before

```
  factory DeviceCalendarPlugin({bool shouldInitTimezone = true}) {
    if (shouldInitTimezone) {
      tz.initializeTimeZones();
    }
    return _instance;
  }
```

andzejsw (approved):

Issue #323, opened 2021-07-27
Samsung Devices move AllDay Events around


albatrosify:
I have gotten a few bug reports, that allDay Events are starting to appear on wrong days (often 1 day earlier than correct).
I did not change anything on my code, neither did this plugin change.
It looks like my app / the plugin is inserting the events correctly, as for 10 second around the elements appear correctly for example in google calendar. After this 10seconds they start to move a day earlier...

Anyone got any similar bug reports or has any idea on how to solve this? I am not finding any issues. iOS is fine, only android in Samsung devices (and i do not own any)

thomassth:
Can you provide more details? Eg.
- what version are you using? Is it still affecting the latest prerelease?
- is it affecting local or synced calendars?
- Since it might be a vendor-specific problem, is there any difference if you view it afterwards via Samsung calendar or Google calendar?

albatrosify:
I have to correct myself. It is no longer only Samsung Devices. The bug exists across devices.
I am using the latest null-safety dev release "^4.0.1-dev.13167"
So far I only know that this issue exists on synced calendars (namely Google Calendar)

I use an iPhone and do not have any problems with sync issues. Most of my users neither have problems even on android devices. 
I tried to reproduce the error on my development pixel 2 xl, as well as AVD Pixel 4 XL.
Everything was fine, UNTIL I changed the timezone in the settings of my google calendar (via web interface).
I switched it to another timezone, looked at the elements (via web interface) and then set it back to the correct time zone.

From this moment on, syncs from android to google calendar via this plugin move ONLY allday events.
Also it appears (when I look at the wrong elements afterwards from my iPhone) that the allday events spans from, for example: 17 Aug -> 16 Aug. This might explain, why the allDay Event shows a day earlier.

thomassth:
Sounds like the problem is about retrieval of events from calendar.

Though I wonder if this is caused by some strange approach handled by Google or bug from the package itself.

We may need to write around the bug if it's the former.

albatrosify:
Is anybody able to reproduce the problem?

FCView:
[...]

Issue #220, opened 2020-04-17
allDay event does not span multiple days


FCView:
I want to be able to create an event that spans several days, and is an All Day event.

Currently, when I try to do so, it only posts to the calendar as a 1-day All Day event on the first day.

```
Event e = new Event(MyCalendarID);
e.eventId = MyEventId;
e.title = "My Title";
e.description = "My description";
e.start = new DateTime.utc(2020, 4, 1);
e.end = new DateTime.utc(2020, 5, 2);
e.allDay = true;
await _deviceCalendarPlugin.createOrUpdateEvent(e);
```

When I later call DeviceCalendarPlugin.retrieveEvents(...), I get the event back, but the end date now matches the start date (Event.start == Event.end) and it displays only on that start date on my device calendar.

This is on Android.

brendabullorini:
My problem is the opposite.
If I put allDay = false, it creates the an all day event

nickrandolph:
Sounds like a bug with the way that we're setting event start and end dates when it's an all day event. Feel free to open a PR with suggested fix

voivoed:
I don't know if anyone already looked into this, but what I'm seeing is this on Android:

1) I created an all day event that goes from 08/18/20 to 08/19/20 (2 full days) on my calendar;
2) when I retrieve the event, the start date is 08/17/20 20:00:00 and end date is 08/18/20 20:00:00;
3) I see the same thing happening on events from the Holiday calendar, start/end date are short 4 hours, making them start on the previous day and end earlier on the actual date;
4) I'm on EST, which is 4 hours earlier than UTC.

Could it be that the plugin treats start/end times for all day as UTC, even though is already on my local timezone, and then converts the dates unnecessarily? That would cause what I'm seeing.

FCView:
@voivoed , see my pull request ( #256 ) in the Example app for what is going on. Basically, when we create the event, it needs to be from midnight to midngiht. However, when we retrieve the allDay event, the Calendar Provider returns it adjusted to our local timezone, as you mentioned. So, we need to adjust it back to UTC to get the correct start and end dates.

My problem was that when using this package to create an allDay event that spans more than one day, this package was only creating it on the first day (it turns out that it was just not setting the end date). That should be fixed in my PR.

voivoed:
@FCView that's what I had to do, I had to convert it back to get the actual dates, then just use the UTC dates as if they were local.

[...]

PR #324, opened 2021-07-31
Enable multiple-day allDay events on Android

FCView:
Replaces my previous PR. Updated to current version and enables multiple-day allDay events on Android. Fix for #220 and also possible fix for #323 as this updates the Android allDay events to correctly be posted at midnight UTC instead of local timezone.

This still doesn't address the issue with local calendars vs synced calendars, as I [mentioned in my comment here.](https://github.com/builttoroam/device_calendar/issues/220#issuecomment-683474498), but I'm not sure that is fixable within the scope of this project.


hyouuu:
Thanks @FCView wish this could be merged soon. Did you figure out any way to fix the issue you mentioned?

FCView:
@hyouuu No, never did find a fix for that other issue. I think that is an issue with Android's Calendar Provider, or the way Android handles local Calendars. I restricted my application to only use synced calendars and haven't had time to go back and really dive deep.

hyouuu:
Thanks for the clarification!

hyouuu:
Hi @FCView encountered a bug while using your branch - if you create any all day events in Android system calendar and then retrieve it, the end date timestamp will be UTC 0:00 which means it will be incorrect in all other timezones

FCView:
@hyouuu, allday events are stored in the Android system as midnight UTC to midnight UTC the following day ([see documentation here](https://developer.android.com/reference/android/provider/CalendarContract.Events#writing-to-events)). The timezone must be in UTC so there is no way to store a different timezone.

Also, storing it in a different timezone causes trouble if the user moves across timezones (as mine do regularly).

I suppose that it could be modified to return the dates in local timezones, but that is just an extra step that I think is best left to the developer based on their use-case. 

I don't remember how the Calendar Provider returns the date, but see the example that I also updated in this branch that shows how to use/display these allday events after retreival:

_formatDateTime() in example/lib/presentation/event_item.dart

thomassth:
The code seems to be working, but code should be changed in `lib` instead of `example`, otherwise everyone needs to copy the implementation on their own.

For example, the end date is fine in events list, but it's incorrect in the datails page because the lib code wasn't fixed.

Also, variables in calendar delegate was changed in between your commits so that need to be changed as well.

FCView:
@thomassth updated with conflict fix and changed implementation to lib instead of example

thomassth (approved):
Tested on my device:

- [x] add and remove multi-day all day events
- [x] cross-checked results with other calendar apps

A proper integration test would be great but this is working as intended. @FCView great job!

PR #344, opened 2021-10-26
Updated pubspec.yaml dependencies to latest

andzejsw:


thomassth:
Seems fine to me

`pedantic` was also discontinued
they are using `lints` now

While you're at it the `analysis_options.yaml` can also be deleted

thomassth (approved):

PR #347, opened 2021-11-01
Using Github Actions

thomassth:
I'm trying to rewrite the current CI/CD using Github Actions

So far CI is simple enough

Things to work on:

- [x] flutter publish test
- [x] generate changelog summary from `CHANGELOG.md`
- [x] add releases to Github

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -10,11 +10,13 @@ on:
     branches: [ develop ]
   pull_request:
     branches: [ develop ]
-
 jobs:
   test:
-    name: Test
-    runs-on: macos-latest
+    name: Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v1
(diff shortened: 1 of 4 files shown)


thomassth:
Seeing other examples, most are recommending utilizing Github's own secret store, and access it during CI/CD

Maybe we should do that as well

MooNag:
I like this. Looks really easy to maintain too. This plugin really needs a better release workflow given how infrequent they are.

thomassth:
All functions should be ready

Some organizing are probably needed; and we may need to test publish on prod because we don't have the cred keys to test

thomassth:
Might want to do a squash merge this time due to all the commits.

sowens-csd (commented):
Are the branches right in prerelease.yml? It says _Development Release_ but then references `release, hotfix`, should that be `develop`? Just trying to make sure I understand the goal of the two Actions. 

Looks like both prerelease and release are set to dry run, not actually publish, right now so they look safe in any event. If that's right then I should approve and we can test. Confirm?

thomassth:
Dry run first just to be sure I think

The branch names and actions came from the original pipeline. not sure why there's that many divisions tho

sowens-csd (approved):
This seems good for testing the dry-run. There's some documentation and other changes but I can contribute after this is merged.

PR #349, opened 2021-11-08
Update README.md

### LINKED ISSUE #348: TZDateTime migration guide
To make the upgrade as painless as possible, we should do a migration guide on TZDateTime.

Once that's done I think we are ready to publish v4 as main release.

thomassth:
Added my own (very similar to example) migration to readme.

Also I don't think I can leave timezone as blank or null.

Closes #348


sowens-csd (approved):
Looks like a good addition.

Issue #346, opened 2021-10-31
CI/CD upgrade


thomassth:
Related to #344

Somehow the latest dependency upgrade is not working here. It seems that the flutter used within the server was and old version. 

We may need to upgrade the current azure pipeline, or maybe switch to github action?

sowens-csd:
Who has access to update the Azure pipeline? Is the pipeline configuration committed with this project? I had a quick look and didn't see it.

sowens-csd:
I'm just catching up to where you were @thomassth, looks like the Azure pipeline is using Flutter 2.2.1. I'll have another look at the Azure pipeline configuration. 

I'd vote for switching to GitHub actions since I'm more familiar with those and I think it would allow all committers better access to the build tooling. I'm assuming that Azure picks up the latest version of the build configuration from the repo but I'm not certain of that. Have you made changes to the Azure build previously?

sowens-csd:
So this issue https://stackoverflow.com/questions/69334583/azure-devops-ci-cd-not-installing-latest-flutter-version suggests that the FlutterInstall task being used is not being maintained anymore. They're suggesting using this one instead https://marketplace.visualstudio.com/items?itemName=hey24sheep.flutter looking for how to do that now.

nickrandolph:
Hi all, I'll try to take a look at this during the week to get the build up and running agin.

thomassth:
I've been playing with github actions on my side

It seems easy enough

sowens-csd:
@thomassth I was going to ask if you'd seen any good news on the Azure build infrastructure but from the look of the commit verification on your recent merge it looks like the answer is no. I was wondering if we should use the work you've put into #347 and close this issue by having commits verified by Actions even if we don't have the full CI/CD stack working yet. Then we could merge #347 and other issues with passing verifications. Then create a new issue to finish off the CI/CD work. 

Thoughts?

thomassth:
I'm working on making dev release automatic but otherwise things are ready to test

Because of how credentials are made, I think we can only test on prod to see if we can publish the package this way.

sowens-csd:
Sounds like great progress, let me know if I can do anything to help. I agree that prod is likely the only way to test publishing.

thomassth:
Ready to test on prod now :)

thomassth:
We should keep this open until we have finished the migration.

thomassth:
I think the keys are currently stored within Azure DevOps' system, not pulled from the build.

Either we
[...]

PR #352, opened 2021-11-10
Updating CI/CD

thomassth:
Continuing #346

- Don't create github releases if it failed to publish
- publication test: Added the flutter environment so credentials are reachable


sowens-csd (approved):
Looking good. The script is pretty readable, nice that I can usually understand what it's doing on a quick review.

PR #354, opened 2021-11-12
Set default Availability to Busy, and minor fixes

thomassth:
- `Availability` now default to busy if you don't have one, also makes it not `required`. 
 ie. Less code need to be edited when upgrading to v4.

- Other formatting and warning fixes as prompted by flutter's checks

Since this benefits migration from previous versions I hope to merge this before v4 main release.


sowens-csd (approved):
Default makes sense to me.

PR #357, opened 2021-11-22
using secrets

thomassth:
Part of #346 

@nickrandolph added the credentials to Github's secrets. (Thanks!)

If this runs we will disable `dry_run` next time.


sowens-csd (approved):
Progress!

PR #359, opened 2021-11-22
final prep

thomassth:
Part of #346 

- Dry-run removed
- Added changelogs to prereleases
- removed pipeline to avoid potential clashes


sowens-csd (commented):
Sorry, this was probably covered elsewhere and I missed it but I wanted to understand the outcome of the prerelease workflow before I approve. 

I see that both prerelease and release have the `dry_run: true` option removed, so both are publishing to pub.dev. So the difference between the two versions is the addition of an explicit version number to the prerelease? In this section I think:

```yaml
name: Edit pubspec.ymal for dev release
        run: |
          sed  -i "s/version.*/&-$GITHUB_RUN_ID/" pubspec.yaml
```

The current prerelease on pub.dev is `4.0.1-dev.13167` will this version keep the same version format? If so I'm not sure where the `-dev.` comes from. Also, on pub.dev currently the published dev version is explicitly listed as a prerelease but I'm not sure how the publish task does that?

thomassth:
According to [this article](https://dart.dev/tools/pub/publishing#publishing-prereleases), if there's anything attached to the end of version number, it is considered to be dev branch. The original pipeline also did it by attaching a build number. 

Right now the github build number is longer than the one from azure, so I didn't put the `dev-` back in.

`GITHUB_RUN_NUMBER` is also avaliable, but it starts with 1 so it may clash with current prerelease. (perhaps we can use this at the next version bump.)

sowens-csd (approved):
Thanks for the explanation, makes sense.

Issue #356, opened 2021-11-17
Version semantics


thomassth:
We are a few buttons away from v4, but actually we have 2 to be pushed according to `changelog.md`

We have
A.
 [4.0.0-dev.12864](https://github.com/builttoroam/device_calendar/releases/tag/4.0.0-dev.12864), which is null safe with a few bug fixes
B. 
dev branch (not on pre-release) using new TZ blocks, among other internal upgrades

I feel that using TZ is a big change, and should be a more significant number bump. Also a null-safe version without TZ support might be favorable for people transitioning to null safety.

Which can be
- A=v4 and B=v4.1
The original plan that v4=null. 
- A=v3.9 and B=v4.0
 People who have `^3.x.x` in their files won't be auto-upgraded to v4, which needs some tinkering to run. But they will get v3.9 that is null safe. 
However people using `^4.0.1-dev` won't get anything until we have v4.0.1 on release. Since we still have a few PRs in the queue I think this can be done soon.

In each case, we can open a temp branch and PR to release.

Depends on #346

sowens-csd:
I'd vote for v4 and v4.1. The null save upgrade to a code base is hard enough, as is testing it. Adding the TZ changes to it will make validating the change harder.

thomassth:
> I'd vote for v4 and v4.1. The null save upgrade to a code base is hard enough, as is testing it. Adding the TZ changes to it will make validating the change harder. 

This is why I have the idea for `3.9`.

If you have `^3.2`, you will be getting `3.9`, with TZ changes. (ie. 4.0.0-12864 = 3.9)

Then, you may opt to upgrade to `^4.0`, with TZ changes etc

sowens-csd:
Your logic makes sense and I'm happy to go whichever way you prefer. 

My thinking, in case it helps, is that the null safety changes are major breaking changes and probably deserving of the 4.x update. I go for more version upgrades rather than less so if the TZ changes are big enough I'd even support 4.x and 5.x, but I know that's not a universal preference.

thomassth:
That's a valid point.

I'm rather confident that null safety doesn't do harm because it seems quite stable and pain-free to me and others here when migrating. Most issues here are either for new features or already present before null safety. 

Let me know if there's some probable issues though, v4 would be more suitable if that's the case.

(v4 and v5 feels kinda too much so I didn't add that in)

sowens-csd:
[...]

PR #361, opened 2021-11-23
V3.9 release

thomassth:
Related to #356 

This is the version from https://github.com/builttoroam/device_calendar/commit/f65738a95e679452ca311b1c41135d6934c8d525, only file versions and pipeline files are changed.

After this is released, we should do a simple push from the dev to main.


thomassth:
Since the pipeline is ready, this is good to go!

sowens-csd (approved):
Although I haven't reviewed these changes this is all historical and we're just trying to get the releases caught up. So, approving on the basis that everyone thought these were okay back when they were done.

Issue #362, opened 2021-11-24
Problems with flutter publish


thomassth:
Part of #346

Appearantly our publish credentials are expired since March. That's why our action stuck in a loop. 

Also see
https://github.com/sakebook/actions-flutter-pub-publisher/issues/22

I think @nickrandolph need to get a new one from pub.dev

sowens-csd:
Thanks @thomassth, I pinged @nickrandolph. He said he'll be online tomorrow morning, hopefully he can resolve then.

sowens-csd:
so @nickrandolph got some new credentials and I tried a build but it still appeared to hang.

thomassth:
It seems that Google needs another layer of auth to proceed. I thought credentials would be enough 😒

Someone need to click the generated link to give auth to GHA access to publish. Probably @nickrandolph 's account for pub.dev

sowens-csd:
Okay, I've pinged Nick.

nickrandolph:
Hi guys. Where do I find the link I need to click through on?

thomassth:
https://github.com/builttoroam/device_calendar/runs/4323788024?check_suite_focus=true

Try the link in the logs

thomassth:
After some more digging apparently we also need to be verified too.

https://github.com/k-paxian/dart-package-publisher/issues/14

(And they've just resolved this less than a month ago. If we did this any sooner we will be totally stuck.)

sowens-csd:
I think the publisher is verified on pub.dev. That means it should work?

sowens-csd:
I killed that task you started @thomassth it had been going for hours. Thoughts?

thomassth:
Forgot to add back the `--force` arg

It asks for confirmation without it.

You can do the edit if you have the time

sowens-csd:
That looks like success @thomassth! I see the new version on pub.dev.

thomassth:
Interestingly `d` goes before numbers

But that's another problem so I'm closing this now!

PR #365, opened 2021-11-25
Redo flutter publish

thomassth:
Part of #362 

Referencing [xvrh's work](https://github.com/xvrh/lottie-flutter).

@nickrandolph also pointed out our json's formatting might have caused problems. Now we wrap it in a file properly so hopefully it works better now.

Also, their pipeline [worked](https://github.com/xvrh/lottie-flutter/runs/3704939536?check_suite_focus=true) so I'm a bit more confident this time around 🤞


sowens-csd (approved):
So we don't need the publish action anymore because the flutter pub publish action will directly use the credentials?

Issue #320, opened 2021-07-20
Json methods use different attributes


CJohnson19:
The `event` model uses attribute names with a different name in `toJson` and `fromJson`. #316 covers the start and end time fields, but `title`, `description`, `allDay`, and `url` also have this issue.

Example:

```dart
Map<String, dynamic> toJson() {
  ...
  data['eventTitle'] = title;
  data['eventDescription'] = description;
```

```dart
Event.fromJson(Map<String, dynamic>? json) {
    ...
    title = json['title'];
    description = json['description'];
```

There is currently a unit test for some of these fields in `Event_Serialises_Correctly`, which fails.

I'd be happy to submit a PR to fix the attribute changes.

I would also suggest having the CI run the `FlutterTest@0` task as a check - and can include that in the same PR if the owners agree it should be included.

thomassth:
We really need more work in test and documentation area, so feel free to make a PR.

thomassth:
#382 should fix any JSONs created after this PR.

Some kind of convertor may be needed for previously created JSONs.

thomassth:
Things should be fixed now. Reopen this if it isn't.

PR #382, opened 2021-12-17
Fixing event JSON mapping

thomassth:
Part of #320

Of course #321 only fixes internal JSON translations.

This fixes JSON translations between app and respective device OS.


sowens-csd (commented):
This was probably covered in some previous PR but why are we changing the names? What's the goal of the change?

thomassth:
It wasn't entirely fixed actually

A lot of event arguments have changed names some time ago, probably during the setup for timezones, and wasn't updated across the board. Ie. Currently the event title and other fields created by `toJSON` can't be read by `fromJson`. Reading and writing to devices were fine though since probably only `toJSON`was edited. 

I only discovered that when I merge the PR for testing.

I just realized that any JSON maps previously created will be unusable. Either we make `toJSON` able to read both of them or we need a mitigation.

thomassth:
This is where it all got wrong

https://github.com/builttoroam/device_calendar/commit/aea337990b410ed411d03bdb80712486195854fd#diff-9e474c05021706fff3750cb85b2d9a25f4d2d528b95c285bb1881cc5bfff03b9

Either we follow through or we need to revert.

sowens-csd (approved):
I tried it on iOS and it seems fine. You said that any maps created previously will be unreadable, you mean anything that was persisted? Within the scope of a single session the values should be consistent now?

thomassth:
> I tried it on iOS and it seems fine. You said that any maps created previously will be unreadable, you mean anything that was persisted? Within the scope of a single session the values should be consistent now?

Any Event JSONs created prior to `4.0` are unreadable after `4.0`. If somebody store event JSONs and reuse them they are in big trouble. 

`4.0` and `4.0.1` will still create faulty JSONs because the conversion is incomplete. Said people are in moderate trouble.

But everything should stay consistent now? We definitely need better testing procedures.

PR #385, opened 2021-12-19
more fixes to integration test

thomassth:
Clearer instruction to run integration test, and fixed test instructions so that the test is actually working.

Also fixed some minor items so the test passes.

iOS should require some other way to grant permission before running the integration test. Like [this](https://stackoverflow.com/questions/66514569/grant-ios-permissions-to-flutter-driver-for-integration-test).

I don't have an iOS device with me so I can't confirm. If it works we can write in our instructions.

### DIFF
--- a/example/integration_test/app_test.dart
+++ b/example/integration_test/app_test.dart
@@ -3,11 +3,10 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:uuid/uuid.dart';
 
-// import 'package:flutter_driver/driver_extension.dart';
 import 'package:device_calendar_example/main.dart' as app;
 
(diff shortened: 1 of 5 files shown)


sowens-csd (commented):
I think there's a fix for the JSON work you were doing mixed in with this commit on the integration test: `event.dart`. If that's right then I think we should keep them separate.

sowens-csd:
I've got a set of iOS integration tests that grant permissions for another project. I'll do a PR with iOS instructions once you're happy with this.

thomassth:
> I think there's a fix for the JSON work you were doing mixed in with this commit on the integration test: `event.dart`. If that's right then I think we should keep them separate. 

You're probably right but I can't fix it now

You can add commits to my PR since you're also a maintainer, if you have the time

sowens-csd (approved):

PR #391, opened 2021-12-21
Change attendance status

thomassth:
Replaces #335. 

cherry picked commit 87710e0f0e084624aac8b477b40d9f1bd42790a1 into the latest branch for convenience.

- [x] Check all iOS variables work
- [x] Finish the `example` viewing, editing and updating attendees
  - [x] Android
  - [x] iOS
- [x] Clear documentations on how iOS and Android implements this

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,13 @@
 # Device Calendar Plugin
 
-**If you're upgrading to v4, your previous code will need to be modified (slightly) otherwise it will not run after update. See [Timezone support](https://github.com/builttoroam/device_calendar#timezone-support-with-tzdatetime) for more details.**
-**There are some changes to event JSON formats at v4. Pay extra care if handld event JSONs. Directly calling to and from device calendars should be unaffected.**
-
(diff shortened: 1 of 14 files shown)


thomassth:
Although the tests are passing, more tests are needed to make sure the function actually works.

- [ ]  Build respective functions into `example` app
- [ ] Clear documentations on how iOS and Android implements this

@maksmaxx do you have any experience using your function?

thomassth:
I've done all I can with an android device. Everything android side are working as intended.

I need someone with an iOS device to complete the rest:

- [ ] Check all iOS variables work
- [ ] Finish the `example` viewing, editing and updating attendees

sowens-csd:
I'll put the iOS work for this next on my list @thomassth.

sowens-csd:
The example continues to work fine on iOS. I can see attendees. What else should I be looking at?

thomassth:
According to the commit, by adding a new IOS only view, you should be able to change attendance details, instead of just viewing them?

@maksmaxx any experience making this work?

sowens-csd:
Okay, I tried out the iOS specific dialog in the example app. A couple of small changes were required to the example app, namely making the Android status nullable and adding a privacy description to the plist. If we want to support this functionality we'll need to change the readme to document this new plist key requirement. The attendee status dialog did work but I think that updating the entry changed the ID of the calendar event which means that the example app has errors after the update until you reload the calendar to get the new event ID. 
@maksmaxx could you comment on whether this is the behaviour you'd expect?

thomassth:
I've updated the plist documents.

@sowens-csd was the attendance update successful? (did you see the updated details at other calendar apps?)
If it does, we only need to update how the `example` app works.

sowens-csd:
So I've just committed an ugly hack to the example that means it doesn't cause an error. It pops back out to the calendar list after editing an iOS entry. With that change the iOS specific behaviour does seem to be working properly.

sowens-csd (approved):

PR #393, opened 2021-12-24
iOS integration test execution framework

sowens-csd:
The project has an Android integration test that does not yet run on iOS. This PR adds the ability to easily execute the test against an iOS device, currently a simulator.


sowens-csd:
@thomassth when you have a chance have a quick look at the approach here. I tried the same approach you took for the Android test of using `Process.run` to execute the `xcrun simctl` logic but I couldn't get it to work. It terminated the test app which causes problems. Using a shell script allowed me to create a simulator for the run then clean up afterwords.

thomassth (approved):

PR #399, opened 2021-12-31
 [Fix][Android] Fix the compile error caused by andoidx.appcompat when building apk in release mode

zhuyangyang-lingoace:
[Fix][Android] Fix the andoidx.appcompat that is incompatibility caused compile error when building apk in release mode

1.error: 'AAPT: error: resource android:attr/lStar' not found when building apk in release mode
2.causes: androidx.appcompat incompatibility

Signed-off-by: zhuyangyang <yangyang.zhu@pplingo.com>

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,7 +50,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'com.google.code.gson:gson:2.8.8'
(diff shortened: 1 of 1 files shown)


thomassth:
Apologies in advance, but I still have trouble comprehending your description. 

Does this solves what you were concerning before (running this plugin in older flutter versions)?

zhuyangyang-lingoace:
@thomassth Sorry. when I run 'flutter build apk',  it compile errors. Just compile example project. So you can build release apk?

thomassth (approved):

PR #410, opened 2022-02-08
Add EventStatus property to Event

lucasribolli:
Android: https://developer.android.com/reference/android/provider/CalendarContract.EventsColumns#STATUS
iOS: https://developer.apple.com/documentation/eventkit/ekevent/1507158-status

This property is useful when the organizer provides a confirmation of the meeting to the attendees. There is more information about it in this [link](https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.11)

In iOS it seems to be read only. That's the reason I add a platform verification in example, allowing EventStatus edition only in Android. But I am not sure if there is a way to edit the status property in iOS side.

I tested it in both platforms


thomassth (approved):

GoldenSoju:
I added the fix to another pull request. ([Link](https://github.com/GoldenSoju/device_calendar/commit/11cb931e6596a402b955924acdc853fc0c6beb26))
It might be in connection with a different recurrence rule implementation.

PR #413, opened 2022-02-20
iOs fix - allDay event does not span multiple days

vasilich6107:


vasilich6107:
Hi @Brett 
I'm fixing this line of your code 
https://github.com/builttoroam/device_calendar/blame/988f671f43516d4f6d7ec4969172b7f4b84fa833/ios/Classes/SwiftDeviceCalendarPlugin.swift#L680

Could you look at my PR

thomassth:
Your fix looked a bit ... simple?

Does it set the multiple all-day events in iOS correctly?

And if that's the case, we need to update the demo app as well

vasilich6107:
Hi @thomassth 
In understand your concerns about the fix)

I will film a short video for you later today.

thomassth:
A simple Y/N would do 😂

And we need to write a test for it too

vasilich6107:
Got you)

Yes it fixes the issue and creates multiday allday event.

I will add the tests and ping you.

thomassth:
You need to rebase your PR before merging.

OleksandraFedotova:
That is not enough for the fix, you do not take into account the timezone set - in your fix - all FullDay events for multiple days and not - will have midnight time - I checked that on the test IOS device
Check  my PR - there is a correct fix for that - https://github.com/builttoroam/device_calendar/pull/450

PR #431, opened 2022-07-20
Allows the plugin to be used in the background / an isolate

lucky-lusa:
CalendarDelegate is always created. In the background without binding. Since the permissions cannot then be queried in the isolate, the check is suspended in this case.


thomassth:

thomassth (approved):

PR #403, opened 2022-01-05
Implementing Rrule package

GoldenSoju:
New Pull Request for #400, related to #298

Currently working in `example`: 

- [x] can build daily or weekly recurring events. 
- [x] can build monthly or yearly events.
- [x] can build complex (eg. multi days a week) events.
- [x] can delete all events in the chain. 
- [x] can delete single events.
- [x] can delete this and following instances

Hurdles (all should be solved before merge):

- [x] `rrule` changes are published.
- [x] most features are working / non-working features hidden and labeled.
- [x] re-enable lower API support (full or partial), or document why they should be dropped.

PR #432, opened 2022-07-30
Fixed bug where DAILY freq gave a compile error

alexthe2:
The `when` branch was not exhaustive enough resulting in build errors when building for Android
```
e: C:\flutter\.pub-cache\hosted\pub.dartlang.org\device_calendar-4.2.0\android\src\main\kotlin\com\builttoroam\devicecalendar\CalendarDelegate.kt: (772, 9): 'when' expression must be exhaustive, add necessary 'DAILY', 'HOURLY', 'MINUTELY', 'SECONDLY' branches or 'else' branch instead

FAILURE: Build failed with an exception.
```

I have added an `else` condition throwing an exception to combat this issue:
```kotlin
else -> throw IllegalStateException("Only Freq WEEKLY, MONTHLY or YEARLY acceptable as a rfcRecurrenceRule!")
```

### DIFF
--- a/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
+++ b/android/src/main/kotlin/com/builttoroam/devicecalendar/CalendarDelegate.kt
@@ -788,7 +788,7 @@ class CalendarDelegate : PluginRegistry.RequestPermissionsResultListener {
                 }?.toMutableList()
             }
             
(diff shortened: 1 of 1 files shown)


alexthe2:
I now also fixed it, and when it's Daily it adds all Days of the week

meunicorn (approved):

thomassth:

thomassth:
Likely solved with #403.

Also not entirely sure what you meant, did not see any build errors like this recently.

PR #442, opened 2022-10-04
updated recurrence frequency

kadariyaujwal:


thomassth:
Recurrence frequency would likely to be reworked with `rrule`. See #403 if you'd like to help.

Issue #427, opened 2022-05-27
PlatformException(500, ownerAccount must not be null, null, null)


SungHyun22:
I installed device_calendar-4.2.0 example app.
and I got this exception on the SM-G960N.

## Flutter doctor

[√] Flutter (Channel stable, 3.0.0, on Microsoft Windows [Version 10.0.19043.1645], locale ko-KR)
• Flutter version 3.0.0 at C:\Users\LEE\flutter
• Upstream repository https://github.com/flutter/flutter.git
• Framework revision ee4e09cce0 (11 days ago), 2022-05-09 16:45:18 -0700
• Engine revision d1b9a6938a
• Dart version 2.17.0
• DevTools version 2.12.2

[√] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
• Android SDK at C:\Users\LEE\AppData\Local\Android\Sdk
• Platform android-Tiramisu, build-tools 30.0.3
• ANDROID_HOME = C:\Users\LEE\AppData\Local\Android\Sdk
• Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
• Java version OpenJDK Runtime Environment (build 11.0.12+7-b1504.28-7817840)
• All Android licenses accepted.

[√] Chrome - develop for the web
• Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop for Windows (Visual Studio Community 2019 16.5.2)
• Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
• Visual Studio Community 2019 version 16.5.29926.136
• Windows 10 SDK version 10.0.18362.0

[√] Android Studio (version 2021.2)
• Android Studio at C:\Program Files\Android\Android Studio
• Flutter plugin can be installed from:
https://plugins.jetbrains.com/plugin/9212-flutter
• Dart plugin can be installed from:
https://plugins.jetbrains.com/plugin/6351-dart
• Java version OpenJDK Runtime Environment (build 11.0.12+7-b1504.28-7817840)

[√] IntelliJ IDEA Community Edition (version 2019.3)
• IntelliJ at C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2019.3.4
• Flutter plugin can be installed from:
https://plugins.jetbrains.com/plugin/9212-flutter
• Dart plugin can be installed from:
https://plugins.jetbrains.com/plugin/6351-dart

[√] VS Code (version 1.67.2)
[...]

PR #443, opened 2022-10-05
Owner account null #427

SungHyun22:
create PR by issue #427


thomassth:
Note: besides fixing #427, I've also include some minor fixes:

- Example app now shows more data on calendar view
- Example app using default dark mode
- Enables flutter_lint; fixed most of the linting issues
- Fixes error "Warning: This version only understands SDK XML versions up to 2 but an SDK XML file of version 3 was encountered." via updating gradle files

thomassth (approved):

Issue #298, opened 2021-05-05
Change of actual RecurrenceRule Class


GoldenSoju:
First of all thanks to everyone for your efforts in 'reviving' the package and the official null-safety version.

I was wondering if someone else would see a benefit in changing the actual RecurrenceRule class to the one from [RRule package (Link)](https://pub.dev/packages/rrule) . The benefit I see, is a better compatibility with external data as the Recurrence Rule can easily converted to an iCalendar/RFC 5545 compliant String (and vice versa). Also some other utility functions like getting recurrences of a recurring event or human-readable text conversion are included.

I am hoping to read your opinions. :-)

andzejsw:
In my opinion this is great idea. I also see that @thomassth agrees. Would like to see what @nickrandolph thinks.
I just think that these changes should be made after PR https://github.com/builttoroam/device_calendar/pull/297 is merged.

nickrandolph:
Yeh I don't see any issues with adopting that package.

GoldenSoju:
Sorry for my late reply, I didn't find earlier time to work on this.

So I tried to implement the RRule package and it worked very well for Android (and Flutter).
(I commited to [my fork](https://github.com/GoldenSoju/device_calendar)).

I wanted to start implementing on iOS today, just to realize that iOS uses the EKRecurrenceRule (EventKit) class and it has no easy way to receive an RFC 5545 String. So while on Android I was able to receive the RFC 5545 String and everything worked just fine, on iOS the EKRecurrenceRule would have to be converted from/to RFC.

Not sure how to proceed from here. Looking at some libraries the conversion EKRecurrenceRule <-> RFC 5545 isn't done in some lines, so that could take me some time implement.

Maybe someone has a better idea?

GoldenSoju:
So I applied the change also to the iOS part. (I commited to [my fork](https://github.com/GoldenSoju/device_calendar)).

For iOS I made use of an exisiting [library](https://github.com/rmaddy/RWMRecurrenceRule). Please note that I am not used to Swift/iOS and didn't manage to import the files correctly. So I copied all relevant code to the top of the `ios/Classes/SwiftDeviceCalendarPlugin.swift` file. Which is super ugly, so it would be great if someone with the ability and time could import the files of the `ios/Classes/EkRruleConverter/*` folder into `SwiftDeviceCalendarPlugin.swift` in a cleaner/more appropriate way, that would be great.
Besides this I had to do changes on the example project to make it run.

[...]

PR #445, opened 2022-10-07
5.0 prerelease

GoldenSoju:
Solves [#298 (Change of actual RecurrenceRule Class)](https://github.com/builttoroam/device_calendar/issues/298).

Breaking Change:
- Replace old Recurrence Rule implementation with [rrule package](https://pub.dev/packages/rrule).

Other Changes:
- Use of Flutter Lints
- Add **event status**
- Several dependency updates
- Sample App updates/improvements
- Other fixes/improvements

Needed before final release:
- Readme/Changelog update

Should we do a prerelease (for one week maybe)?


thomassth:
I was thinking 4.3 but that works too

I think we can do 2 weeks prerelease just to be safe

Also, I was thinking about adding some compatibility to the old recurrence rule, to make the switchover easier

PR #449, opened 2022-10-12
packages cleanup

thomassth:
No more hacky `timezone` imports needed, now that the files are (correctly) exported.

Also removed packages that ain't essential, slimming it down

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,10 +7,7 @@ dependencies:
   flutter:
     sdk: flutter
   collection: ^1.16.0
-  sprintf: ^6.0.2
   timezone: ^0.9.0
-  flutter_native_timezone: ^2.0.0
-  intl: ^0.17.0
   rrule: ^0.2.7
 
 dev_dependencies:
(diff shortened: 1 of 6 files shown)


pubspec.yaml
```diff
@@ -7,10 +7,7 @@ dependencies:
   flutter:
     sdk: flutter
   collection: ^1.16.0
-  sprintf: ^6.0.2
   timezone: ^0.9.0
-  flutter_native_timezone: ^2.0.0
-  intl: ^0.17.0
```
IVLIVS-III, line 13: are these three dependencies not used anymore or why can they be removed?

thomassth: These are used in the example app, but not really used in the package itself.

IVLIVS-III: Shouldn't we add them as dependencies to the `example/pubspec.yaml` then?

Maybe I misunderstood, but as far as I can tell, now the example app cannot access those three packages anymore (haven't tested the example app tho).

thomassth: They're already there lol
No one bothered to clean it up before I think

IVLIVS-III: You're right, `intl` and `flutter_native_timezone` are already there. `sprintf` is missing, but never imported, thus not necessary.

Issue #337, opened 2021-10-06
All day events start time may be incorrect (iOS only)


scorpionate:
I noticed  **startTimeZone: ekEvent.timeZone?.identifier**, might be null(nil) for all-day events.
Because of that, in my flutter code after calling _retrieveEvents()_ I received an all-day event object with changed start time property.

Let's say my event is at 06.10.2021 and my timezone is +2.
I got an event object with 
- start time at 05.10.2021, 22:00 (timezone difference exactly)
- end time at 06.10.2021, 24:00

SwiftDeviceCalendarPlugin.dart
```
 private func createEventFromEkEvent(calendarId: String, ekEvent: EKEvent) -> Event {
        ...
        ...
        let event = Event(
            eventId: ekEvent.eventIdentifier,
            calendarId: calendarId,
            title: ekEvent.title ?? "New Event",
            description: ekEvent.notes,
            start: Int64(ekEvent.startDate.millisecondsSinceEpoch),
            end: Int64(ekEvent.endDate.millisecondsSinceEpoch),
            startTimeZone: ekEvent.timeZone?.identifier,
            allDay: ekEvent.isAllDay,
            attendees: attendees,
            location: ekEvent.location,
            url: ekEvent.url?.absoluteString,
            recurrenceRule: recurrenceRule,
            organizer: convertEkParticipantToAttendee(ekParticipant: ekEvent.organizer),
            reminders: reminders,
            availability: convertEkEventAvailability(ekEventAvailability: ekEvent.availability)
        )

        return event
    }
```

thomassth:
Can you try again with the latest build in #385. I think things should be fixed but I don't have an iOS device to confirm.

OleksandraFedotova:
there is a broken part here https://github.com/builttoroam/device_calendar/blob/develop/ios/Classes/SwiftDeviceCalendarPlugin.swift
You set endDate to startDate if AllDay is true 
check the 818 line 
    if (isAllDay) { ekEvent!.endDate = startDate }
            else {
                ekEvent!.endDate = endDate

OleksandraFedotova:
I added PR https://github.com/builttoroam/device_calendar/pull/450 that fixes that problem

henry2man:
Can we consider this issue as fixed?

IVLIVS-III:
@henry2man yes, good catch. The PR was merged in the 4.3 release.
Thanks for the reminder.

PR #450, opened 2022-10-13
Fixed All day events start time may be incorrect (iOS only) #337

OleksandraFedotova:
Fixed fullDay with DateRange events creation (not only one fullday events possible to create) Timezone should be set only if event is not full day

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -815,12 +815,12 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
             ekEvent!.notes = description
             ekEvent!.isAllDay = isAllDay
             ekEvent!.startDate = startDate
-			ekEvent!.endDate = endDate
+            ekEvent!.endDate = endDate
             
 			if (!isAllDay) { 
-				let timeZone = TimeZone(identifier: startTimeZoneString ?? TimeZone.current.identifier) ?? .current
+                let timeZone = TimeZone(identifier: startTimeZoneString ?? TimeZone.current.identifier) ?? .current
                 ekEvent!.timeZone = timeZone
-			}
+            }
             
             ekEvent!.calendar = ekCalendar!
             ekEvent!.location = location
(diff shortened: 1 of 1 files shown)


ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -815,13 +815,13 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
-
-                let timeZone = TimeZone(identifier: startTimeZoneString ?? TimeZone.current.identifier) ?? .current
+			ekEvent!.endDate = endDate
+            
+			if (!isAllDay) { 
+				let timeZone = TimeZone(identifier: startTimeZoneString ?? TimeZone.current.identifier) ?? .current
                 ekEvent!.timeZone = timeZone
-            }
+			}
+
```
IVLIVS-III, line 824: The formatting seems a bit off, looks like the new lines have double the indentation. Please make sure to use spaces instead of tabs.

OleksandraFedotova: Fixed formatting to spaces

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -815,13 +815,13 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
             ekEvent!.notes = description
             ekEvent!.isAllDay = isAllDay
             ekEvent!.startDate = startDate
-            if (isAllDay) { ekEvent!.endDate = startDate }
-            else {
-                ekEvent!.endDate = endDate
-
+            ekEvent!.endDate = endDate
+            
+			if (!isAllDay) {
```
IVLIVS-III, line 820: sorry that I am meticulous here… line 820 still has tabs

IVLIVS-III (changes requested):

IVLIVS-III:
Could you please describe what exactly this PR fixes.
What went wrong / didn't work as expected?

It looks like this PR also introduces all-day-multi-day events. Is that correct?

IVLIVS-III:
I just tested your changes on iOS.
Works great, all-day-multi-day events can be read from the calendar without issues.

Could you please also change the example app, so when toggling all-day to on, the input field of event end no longer disappears.
With your PR hiding this field is no longer necessary on iOS.
It would be good, if we could also test creating multi-day dll-day events from within the example app.

IVLIVS-III:
Great, that's what I thought. Is a very welcome change.

Perfect if you can change the example app as well, thank you.

IVLIVS-III:
Since I don't want this PR to be blocked any longer, waiting only for changes in the example app, I split it up into a separate PR (see https://github.com/builttoroam/device_calendar/pull/454).

Thanks @OleksandraFedotova for your contribution.

IVLIVS-III (approved):
LGTM

PR #453, opened 2022-11-01
iOS: Fix error when adding attendees due to UUID is nil.

GoldenSoju:
I recently had several crashes on my app (iOS users) and found out that the adding attendee functionality was broken.

#### Error:
```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSDictionaryM setObject:forKeyedSubscript:]: key cannot be nil'
```

I was able to locate the error in the _SwiftDeviceCalendarPlugin.swift_ file (line 692):

```swift
ekEvent!.setValue(attendees, forKey: "attendees")
```

And adding the line below to the createParticipant in the same file solved the problem.
*Thanks to this [solution](https://stackoverflow.com/a/74265315/13558952).*
```swift
participant.setValue(UUID().uuidString, forKey: "UUID")
```

Besides that I

- had to change some null errors in the example app (/event_attendee.dart),
- changed a color value for dark mode in (/calendar_event.dart) as the greenAccent[100] looks awful in dark mode. ;-),
- updated rrule package dependency from 0.2.7 to 0.2.10,
- deleted unneccessary dependency in _test/device_calendar_test.dart_

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -25,10 +25,10 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   device_calendar: 9cb33f88a02e19652ec7b8b122ca778f751b1f7b
-  Flutter: 50d75fe2f02b26cc09d224853bb45737f8b3214a
+  Flutter: f04841e97a9d0b0a8025694d0796dd46242b2854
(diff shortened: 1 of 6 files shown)


IVLIVS-III:
Looks good on first glance, thanks. Will hopefully have time to do a proper review in the coming days.

IVLIVS-III (approved):
LGTM

PR #465, opened 2022-12-31
Remove faulty requestPermissions implementation in favor of working one

tjarvstrand:
Tthe current Swift-implentation of `requestPermissions` does not return if permissions have already been granted. Thus, if it is called more than once, subsequent attempts try to request permissions again, causing the call to fail with a 400-error (`To create or update an all day event you must provide calendar ID, event with a title and event's start date`).

There was already a second implementation of `requestPermissions` which wasn't used anywhere but correctly short-circuits if permissions have already been granted so I thought it would be better to use that.

EDIT: Actually, the errors seem to be caused by an error on my end. Regardless, this change seems reasonable to me, if less urgent :)

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -1024,7 +1024,7 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
         self.finishWithUnauthorizedError(result: result)
     }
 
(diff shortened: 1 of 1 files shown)


IVLIVS-III:
Thanks for the PR, looks good on first glance. I will try and test it soon.

IVLIVS-III (approved):
LGTM

IVLIVS-III:
Thank you for your contribution @tjarvstrand

PR #466, opened 2022-12-31
macOS Supported.

kingvhit:
I have modified from iOS Source Code and make it work in macOS.

What I have changed:
- Reuse model attendee from iOS. (dart file)
- Update Plugin configuration.
- Modified DeviceCalendarPlugin swift file, I think we can re-use if some one has more exprience in Swift (not by me)
+ Remove UIKit, EventKitUI (only supported in iOS). 
+ Remove show event detail in modal by iOS (macOS does not has UIKit).
+ Migrate from UIColor -> NSColor (supported by macOS)

---
For the future, I think we can conduct code in Swift and find the extension for support both iOS and macOS.

Feel free ask me if you want to do anything.

---
Notes:
- Please add this into Info.plist
```
	<key>NSCalendarsUsageDescription</key>
	<string>Access most functions for calendar viewing and editing.</string>
```
- Also, add this into entitlement file
```
	<key>com.apple.security.personal-information.calendars</key>
	<true/>
```

In-case you run code for checking, please run by command-line instead of dirrectly from VS code or Android Studio, I think that bug from Flutter.

```
flutter run -d macos
```


IVLIVS-III:
On first glance, it looks like you have copied the source file. From your description it seems like we could share most of the code for iOS and macOS.
In that case a file copy duplicating most platform-specific code is not the preferred solution since that would require us to maintain two mostly duplicate code bases in the future instead of a shared one.

I have not looked into your PR in detail yet. Even with the file copy it appears to be a promising proof of concept.

You could look at [this](https://github.com/flutter/plugins/pull/6517) on how code-reuse on iOS and macOS might look like. I sadly don’t have enough time at the moment to do the reuse refactor myself.

Thank you for your effort.

kingvhit:
To refactor the iOS and macOS source code, I think that take huge time todo it. And that need the person in charge who has more exprience for both iOS and macOS development.

By the way, I am not actually expert in Swift, refactor it might be difficult from my side. So, based on your feedback, I will close this PR and it should be maintained by some one in future.

Thank you for your feedbacks.

Issue #452, opened 2022-11-01
Use the retrieveEvents method in the ios device to query the events of the system calendar


w9623:
**Describe the bug**
When I retrieve the schedule of the system calendar using an ios device, I call the retrieveEvents method, set the start time DateTime(1971, 1,1), DateTime(2049, 12,31), and the query result is empty

**To Reproduce**
1.Creates a schedule in the device's system calendar with a schedule time between the start time of the call parameter.
2.Here's the code
```
final calendarsResult = await _deviceCalendarPlugin.retrieveCalendars();

      var _calendars = calendarsResult.data as List<Calendar>;
      for (var _calendar in _calendars) {
        var calendarEventsResult = await _deviceCalendarPlugin.retrieveEvents(
          _calendar.id,
          RetrieveEventsParams(
            startDate: DateTime(1971, 1, 1),
            endDate: DateTime(2049, 12, 31),
          ),
        );
        var _calendarEvents = calendarEventsResult.data as List<Event>;
      }
```
3._calendarEvents is empty

**Expected behavior**
It seems that the start time needs to be set to 2019 before the query succeeds. I need to set a slightly distant historical time.

**Screenshots**

**Device(s) tested**
This can be very important as not all device vendors do calendar in the same way.
 - Device: [iPhone 13mini]
 - OS: [ios 16.0]
 - Plugin version [4.2.0 Release]

**Flutter doctor**

Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.0.5, on macOS 12.6.1 21G217 darwin-x64, locale zh-Hans-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.0.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] Connected device (4 available)
[✓] HTTP Host Availability

**Additional context**

IVLIVS-III:
Without looking into this on the code-level, I can confirm from personal experience, that the retrieved List of Events is empty if the time range is too big.

It might be possible to query for past Events (even those in the distant past) by specifying a smaller time range, that also has the end date in the past.
I didn't have time to test this workaround tho.

@w9623 in case you try this proposed potential workaround, could you please reply with the results. This would help us to find a solution for this issue.

w9623:
> 

I tried to start before 2019 and the time range is very small, but I still can't retrieve the results

[...]

PR #486, opened 2023-05-13
Fix retrieving events for more than 4 years range

VladyslavBilomeria:
Fix for https://github.com/builttoroam/device_calendar/issues/452

The current implementation of receiving events on iOS does not take into account that `EKEventStore` returns events for a maximum of 4 years ([doc](https://developer.apple.com/documentation/eventkit/ekeventstore/1507479-predicateforevents#discussion)). Accordingly, when specifying a range of more than 4 years, not all events are received.

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -329,10 +329,6 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
                         // Adding 4 years to the start date
                         var currentEndDate = startDate.addingTimeInterval(fourYearsTimeInterval)
                         while currentEndDate <= endDate {
-                            // debugPrint("Start date of current range: \(currentStartDate)")
-                            // debugPrint("End date of current range: \(currentEndDate.addingTimeInterval(-1))")
-                            // debugPrint("Range size: \(roundedRangeSize / (365 * 24 * 60 * 60)) years\n")
-                            
                             let predicate = self.eventStore.predicateForEvents(
                                 withStart: currentStartDate,
                                 end: currentEndDate.addingTimeInterval(-1),
@@ -347,12 +343,6 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
                         
                         // If the cycle doesn't end exactly on the end date
                         if currentStartDate <= endDate {
-                            let finalRangeSize = endDate.timeIntervalSince(currentStartDate)
-                            
-                            // debugPrint("Start date of final range: \(currentStartDate)")
-                            // debugPrint("End date of final range: \(endDate)")
-                            // debugPrint("Range size: \(finalRangeSize / (365 * 24 * 60 * 60)) years\n")
-                            
                             let predicate = self.eventStore.predicateForEvents(
                                 withStart: currentStartDate,
                                 end: endDate,
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
                     let endDate = Date (timeIntervalSince1970: endDateDateMillisecondsSinceEpoch!.doubleValue / 1000.0)
                     let ekCalendar = self.eventStore.calendar(withIdentifier: calendarId)
                     if ekCalendar != nil {
-                        let predicate = self.eventStore.predicateForEvents(
-                            withStart: startDate,
-                            end: endDate,
-                            calendars: [ekCalendar!])
-                        let ekEvents = self.eventStore.events(matching: predicate)
+                        var ekEvents = [EKEvent]()
+                        let fourYearsInSeconds = 4 * 365 * 24 * 60 * 60
+                        var currentStartDate = startDate
+                        // Adding 4 years to the start date
+                        var currentEndDate = startDate.addingTimeInterval(TimeInterval(fourYearsInSeconds))
+                        while currentEndDate <= endDate {
+                            let rangeSize = currentEndDate.timeIntervalSince(currentStartDate)
+                            let roundedRangeSize = Int(rangeSize / Double(fourYearsInSeconds)) * fourYearsInSeconds
+                            
+                            // debugPrint("Start date of current range: \(currentStartDate)")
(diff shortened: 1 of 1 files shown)


VladyslavBilomeria:
@thomassth Could you please take a look at this when you have time?

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
+                        var currentStartDate = startDate
+                        // Adding 4 years to the start date
+                        var currentEndDate = startDate.addingTimeInterval(TimeInterval(fourYearsInSeconds))
+                        while currentEndDate <= endDate {
+                            let rangeSize = currentEndDate.timeIntervalSince(currentStartDate)
+                            let roundedRangeSize = Int(rangeSize / Double(fourYearsInSeconds)) * fourYearsInSeconds
+                            
+                            // debugPrint("Start date of current range: \(currentStartDate)")
+                            // debugPrint("End date of current range: \(currentEndDate.addingTimeInterval(-1))")
+                            // debugPrint("Range size: \(roundedRangeSize / (365 * 24 * 60 * 60)) years\n")
```
IVLIVS-III, line 336: please completely remove (not just comment out) debug print statements

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
+                            currentEndDate = currentStartDate.addingTimeInterval(TimeInterval(roundedRangeSize))
+                        }
+                        
+                        // If the cycle doesn't end exactly on the end date
+                        if currentStartDate <= endDate {
+                            let finalRangeSize = endDate.timeIntervalSince(currentStartDate)
+                            
+                            // debugPrint("Start date of final range: \(currentStartDate)")
+                            // debugPrint("End date of final range: \(endDate)")
+                            // debugPrint("Range size: \(finalRangeSize / (365 * 24 * 60 * 60)) years\n")
```
IVLIVS-III, line 356: please completely remove (not just comment out) debug print statements

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
+                            ekEvents.append(contentsOf: batch)
+                            
+                            // Move the start and end dates forward by the rounded range size
+                            currentStartDate = currentEndDate
+                            currentEndDate = currentStartDate.addingTimeInterval(TimeInterval(roundedRangeSize))
+                        }
+                        
+                        // If the cycle doesn't end exactly on the end date
+                        if currentStartDate <= endDate {
+                            let finalRangeSize = endDate.timeIntervalSince(currentStartDate)
```
IVLIVS-III, line 352: This value is never used, thus we could remove calculating it.

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
-                            calendars: [ekCalendar!])
-                        let ekEvents = self.eventStore.events(matching: predicate)
+                        var ekEvents = [EKEvent]()
+                        let fourYearsInSeconds = 4 * 365 * 24 * 60 * 60
+                        var currentStartDate = startDate
+                        // Adding 4 years to the start date
+                        var currentEndDate = startDate.addingTimeInterval(TimeInterval(fourYearsInSeconds))
+                        while currentEndDate <= endDate {
+                            let rangeSize = currentEndDate.timeIntervalSince(currentStartDate)
+                            let roundedRangeSize = Int(rangeSize / Double(fourYearsInSeconds)) * fourYearsInSeconds
```
IVLIVS-III, line 332: Why do we need to round the `rangeSize`?

As far as I can tell, we only use the `roundedRangeSize` to advance the start / end dates for the next batch. Why can't we just use `fourYearsInSeconds` instead?

And if the `roundedRangeSize` is needed, why do we need to calculate it every iteration?

ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -322,11 +322,47 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
...
+                            let predicate = self.eventStore.predicateForEvents(
+                                withStart: currentStartDate,
+                                end: currentEndDate.addingTimeInterval(-1),
+                                calendars: [ekCalendar!])
+                            let batch = self.eventStore.events(matching: predicate)
+                            ekEvents.append(contentsOf: batch)
+                            
+                            // Move the start and end dates forward by the rounded range size
+                            currentStartDate = currentEndDate
+                            currentEndDate = currentStartDate.addingTimeInterval(TimeInterval(roundedRangeSize))
```
IVLIVS-III, line 347: Can't we just use `fourYearsInSeconds` instead of `roundedRangeSize`?

IVLIVS-III (changes requested):
Overall it looks great, thank you very much for addressing this issue.

I just have some minor questions that should be discussed before merging. Please respond to the code-review comments directly.

ios/Classes/SwiftDeviceCalendarPlugin.swift
VladyslavBilomeria: Done

ios/Classes/SwiftDeviceCalendarPlugin.swift
VladyslavBilomeria: Done

ios/Classes/SwiftDeviceCalendarPlugin.swift
VladyslavBilomeria: This was calculated only for logs. Removed

ios/Classes/SwiftDeviceCalendarPlugin.swift
VladyslavBilomeria: I overcomplicated the calculations a little, it really could have been done much easier 😅
Updated

IVLIVS-III (approved):
LGTM

IVLIVS-III:
Thank you for contributing your proposed fix of #452 @VladyslavBilomeria.

I sadly cannot specify an ETA of when this fix will be published on pub.dev, since development of this plugin is really slow at the moment.

PR #516, opened 2023-11-28
Added event color field for Android

NaoKreuzeder:
Now able to get the event-color from the Event object


Peng1104:
Feature request from: #517

IVLIVS-III (changes requested):
Please remove the code thats commented out, e.g. `// name …`.

Also, is there a similar concept of event color on iOS? If yes, then we should implement the feature on both platforms.

chris-wolf:
Created a new PR: https://github.com/builttoroam/device_calendar/pull/546 which includes the loading logic of this PR and adds updateEventColor functionality

PR #531, opened 2024-03-17
Fix: Update Set<T> to List<T> in calendar_event.dart to comply with rrule 0.2.16

sheng-di:
This PR addresses the type declaration change in the `rrule` library version 0.2.16. The library now uses `List<T>` instead of `Set<T>`, which necessitates the same change in `calendar_event.dart` of the `builttoroam/device_calendar` package.

Changes made:

- All instances of `Set<T>` in `calendar_event.dart` have been updated to `List<T>`.
- Necessary adjustments in methods/functions that were using `Set<T>` are made to handle `List<T>`.

This change ensures compatibility with the `rrule` library version 0.2.16 and prevents possible type errors. It is recommended to merge this PR to maintain the integrity and functionality of the `builttoroam/device_calendar` package.

Please review and let me know if any changes are required.

PR #513, opened 2023-11-07
WIP: Pipeline update

thomassth:
Uses the official pipeline instead of manual tokens
ie. no more expired token roadblocks
https://dart.dev/tools/pub/automated-publishing

PR #532, opened 2024-03-18
Fix iOS workflow.

IVLIVS-III:
XCode version 15.0-beta is no longer available and causes all tests to break. Use version 15.0.1 instead.

This is a blocker for #531.


IVLIVS-III:
Note: Tests here are failing later in the workflow, which would then be fixed by #531.

IVLIVS-III:
obsolete after #513.

PR #552, opened 2024-08-02
Make the EKEventStore instance a Singleton

### LINKED ISSUE #551: Singleton pattern on EKEventStore instance
Hello,

I'm using device_calendar through a singleton custom class CalendarService and I've been experiencing issues on iOS 17 lately.
Every time I wait a few seconds before approving calendar permission, the latter seems not to be taken into account until I restart my app.

```console
INVOKE_CHANNEL_METHOD_ERROR: PlatformException(401, The user has not allowed this application to modify their calendar(s), null, null)
```

I went through EventKit documentation and I saw that EKEventStore instances should be long-lived : created and destroyed as little as possible during the lifetime of an application instance. So I tried implementing the Singleton pattern on our EKEventStore instance and it solved my problem. Which does not make much sense to me as I've got the same instance of DeviceCalendarPlugin everywhere in my app, so the instance of EKEventStore is the same.

```console
# instance printed before requesting permission
flutter: <EKEventStore: 0x7febac45e680>
# instance printed after permission and before retrieving calendars
flutter: <EKEventStore: 0x7febac45e680>
```

 I don't understand why it behaves this way. My only guess is that the static method `EKEventStore.authorizationStatus(for: .event)` does not catch up fast enough after eventStore request is approved, but it would be weird.
If anyone has a clue, I'd love to hear from you.

Should we make the EKEventStore instance a singleton ?

Thank you

AlexisChoupault:
Somehow fixes https://github.com/builttoroam/device_calendar/issues/551

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -23,6 +23,16 @@ extension String {
     }
 }
 
+fileprivate class SingleEventStore {
+    public let eventStore: EKEventStore
+    
+    public static let shared = SingleEventStore()
+
+    private init() {
+        eventStore = EKEventStore()
+    }
+}
+
 public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDelegate, UINavigationControllerDelegate {
     struct DeviceCalendar: Codable {
         let id: String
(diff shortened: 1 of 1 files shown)


ios/Classes/SwiftDeviceCalendarPlugin.swift
```diff
@@ -23,6 +23,16 @@ extension String {
...
 }
 
+fileprivate class SingleEventStore {
+    public let eventStore: EKEventStore
+    
+    public static let shared = SingleEventStore()
+
+    private init() {
+        eventStore = EKEventStore()
+    }
```
dsmailes, line 33: Doesn't this initialise two instances of EKEventStore? 

1: `eventStore` which doesn't appear to be used, and 2: `shared`

AlexisChoupault: I don't think so. The whole point of the Singleton pattern is that the class and its attributes are instantiated only once at runtime.
In my code the only way to instantiate a `SingleEventStore` is through its static instance `shared`, which calls the constructor where `eventStore` is instantiated (only once)

Issue #561, opened 2024-10-08
iOS 17 PlatformException(401, The user has not allowed this application to modify their calendar(s), null, null)


FySanta:
ios 17.0  
var calendarsResult = await _deviceCalendarPlugin.retrieveCalendars() has error PlatformException(401, The user has not allowed this application to modify their calendar(s), null, null)

   await _deviceCalendarPlugin.requestPermissions();
   await _deviceCalendarPlugin.hasPermissions();
   var calendarsResult = await _deviceCalendarPlugin.retrieveCalendars();

![31728382804_ pic](https://github.com/user-attachments/assets/6d486d19-acf0-4c38-827a-5eebc8253f32)
errorcode：599
errorMessage：Device calendar plugin ran into an issue. Platform specific exception [401], with message :"The user has not allowed this application to modify their calendar(s)", has been thrown.

When getting the fullaccess privileges for a calendar, running await _deviceCalendarPlugin.requestPermissions() gives true after the user grants the privileges, but await _deviceCalendarPlugin.hasPermissions() says is false, and only becomes true after restarting the app.

IVLIVS-III:
make sure that this line of code `await _deviceCalendarPlugin.hasPermissions();` returns `true`. Otherwise permissions have not been granted.

FySanta:
> 确保此行代码`await _deviceCalendarPlugin.hasPermissions();`返回`true`。否则权限未授予。

It shows authorized inside the app settings, but `await _deviceCalendarPlugin.hasPermissions()` return false

FySanta:
![91728439398_ pic](https://github.com/user-attachments/assets/7dcb1909-4bac-40c5-adfc-63b731be52fa)

IVLIVS-III:
That is strange, can you provide a minimum example to reproduce this issue?

FySanta:
Here's my running order
`    await _deviceCalendarPlugin.requestPermissions();
    await _deviceCalendarPlugin.hasPermissions();
    var calendarsResult = await _deviceCalendarPlugin.retrieveCalendars();`

FySanta:
> That is strange, can you provide a minimum example to reproduce this issue?

Here's my running order
```
await _deviceCalendarPlugin.requestPermissions(); 
await _deviceCalendarPlugin.hasPermissions(); 
var calendarsResult = await _deviceCalendarPlugin.retrieveCalendars();
```
You can check separately the permission status returned by these two lines of code.
```
[...]

PR #578, opened 2025-01-10
Fix for event permissions returning false in iOS17+

gavira:
Might fix error when checking for permissions in iOS17/iOS18

Related bug: https://github.com/builttoroam/device_calendar/issues/561
Reference: https://forums.developer.apple.com/forums/thread/739315

### DIFF
--- a/ios/Classes/SwiftDeviceCalendarPlugin.swift
+++ b/ios/Classes/SwiftDeviceCalendarPlugin.swift
@@ -1104,6 +1104,7 @@ public class SwiftDeviceCalendarPlugin: NSObject, FlutterPlugin, EKEventViewDele
         }
     }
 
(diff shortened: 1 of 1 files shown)


IVLIVS-III (approved):
LGTM

IVLIVS-III:
looking at the failing test, the fix requires iOS 13.0 or later.
We could either introduce a version guard or maybe even drop support for iOS < 13.0 entirely.

Anyone wanting to share their thoughts on the topic?

gavira:
Ah sorry, I'm targetting iOS 15 in my app indeed. As far as I know all iPhones that supported iOS 13 also support iOS 14 and 15 so I'd say it's probably safe to drop support?

PR #593, opened 2025-04-04
Fix recurring event deletion by clearing recurrence fields

chris-wolf:
This pull request addresses an issue where deleting a recurring event on local calendars was corrupting the instance projection, causing unrelated events to disappear from queries. The underlying problem was that deleting the event left orphaned recurrence data (such as RRULE, EXRULE, and EXDATE) in the calendar provider, which interfered with how the Instances table was expanded.

Changes Made:

Clearing Recurrence Fields: Before deleting a recurring event, the code now updates the event to set the recurrence-related fields (RRULE, EXRULE, EXDATE) to null. This prevents leftover recurrence data from affecting the instance projection.

Old behavior:
Deleting the recurring event makes the single event dissapear
https://github.com/user-attachments/assets/cdc164e9-3f3f-420d-87d0-08ab73ec73e5

New behavior:
Deleting the reccuring event has no effect on the single event
https://github.com/user-attachments/assets/d62c6b27-a105-46db-8b2c-33b733f18f56


chris-wolf:
Issue seems to still not be fixed.
A fix can be to set
values.put(CalendarContract.Calendars.SYNC_EVENTS, 1)
when creating events.
That causes for events to only be soft deleted for local device calendar and fixed the disappearing of events. but will test further before suggesting a pull request since that would be a weird solution.

PR #594, opened 2025-04-10
Use full URL and not data URI for events

### LINKED ISSUE #574: Data URI instead of normal URI?
Hey, using the latest `device_calendar`, I'm trying to set the `url` param on `Event` with a URL directing to my custom deeplink, but it's not saved in calendar, I have no URL there. So I look into the plugin's code and I see this:

https://github.com/builttoroam/device_calendar/blob/6a9733c35fb693a5fad93ad641886e393a64de11/lib/src/models/event.dart#L246

Why would you accept only [`data:` URIs](https://api.flutter.dev/flutter/dart-core/Uri/data.html)? Should that be fixed and `toString()` any `Uri`s? Do you accept PR for that?

Albert221:
Resolves #574.

Tested on iOS and sets the URL in the calendar nicely.

### DIFF
--- a/lib/src/models/event.dart
+++ b/lib/src/models/event.dart
@@ -157,7 +157,7 @@ class Event {
     if (foundUrl?.isEmpty ?? true) {
       url = null;
     } else {
-      url = Uri.dataFromString(foundUrl as String);
+      url = Uri.tryParse(foundUrl as String);
     }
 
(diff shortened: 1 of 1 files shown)


Albert221:
Hello, @thomassth? The CI fail is not related to my changes BTW

Albert221:
hello

